PR #11, opened 2016-11-04
Add ability to optimize function calls with lambda in function position. Use it to optimize jss. <http://abcl.org/trac/ticket/420>

alanruttenberg:
Precompiler: When compiling a form with a lambda in the function
position, possibly optimize it

Case 1: If the lambda has a single form in it, let someone define a
transform using:

define-function-position-lambda-transform (body-function-name (arglist form args) &body body)

body-function-name is the car of the single form in the lambda
arglist is the arglist of the lambda
form is the single form within the lambda
args are the arguments to which the lambda will be defined.

The function should check whether it can do a transform, and do it if
so, otherwise return nil signalling it couldn't

Case 2: If case 1 is not successful then if the arglist is a simple
one (no &key, &rest, &optional) then do a standard beta-reduction
binding the args to arglist using let (https://wiki.haskell.org/Beta_reduction)

If not, return and do the usual thing.

An example is in contrib/jss/optimize-java-call.lisp

To see benefits, (compile-file contrib/jss/test-optimize-java-call.lisp)
and then load the compiled file. You should see something like the below
which reports the timings for the optimized and unoptimized version of
10000 calls of (#"compile" 'regex.pattern ".*")

--

With optimization: (INVOKE-RESTARGS-MACRO "compile" (QUOTE REGEX.PATTERN) (LIST ".*") NIL T)
Without optimization: ((LAMBDA (#:G85648 &REST #:G85649) (INVOKE-RESTARGS "compile" #:G85648 #:G85649 NIL)) (QUOTE REGEX.PATTERN) ".*")

JUST-LOOP
0.0 seconds real time
0 cons cells

OPTIMIZED-JSS
0.011 seconds real time
0 cons cells

UNOPTIMIZED-JSS
0.325 seconds real time
800156 cons cells

See:
https://mailman.common-lisp.net/pipermail/armedbear-devel/2016-October/003726.html
https://mailman.common-lisp.net/pipermail/armedbear-devel/2016-November/003733.html

### DIFF
--- a/contrib/jss/invoke.lisp
+++ b/contrib/jss/invoke.lisp
@@ -286,8 +286,8 @@ want to avoid the overhead of the dynamic dispatch."
                               (group "group"))
       (loop while (hasmore entries)
          for name =  (getname (next entries))
-         with class-pattern = (#"compile" '|java.util.regex.Pattern| ".*\\.class$")
-         with name-pattern = (#"compile" '|java.util.regex.Pattern| ".*?([^.]*)$")
+         with class-pattern = (jstatic "compile" "java.util.regex.Pattern" ".*\\.class$")
+         with name-pattern = (jstatic "compile" "java.util.regex.Pattern" ".*?([^.]*)$")
          when (matches (matcher class-pattern name))
          collect
            (let* ((fullname (substring (jreplace name #\/ #\.) 0 (- (jlength name) 6)))
--- a/contrib/jss/invoke.lisp
+++ b/contrib/jss/invoke.lisp
@@ -123,6 +123,8 @@
   (defvar *do-auto-imports* t 
     "Whether to automatically introspect all Java classes on the classpath when JSS is loaded."))
 
+(defvar *muffle-warnings* t) 
+
 (defvar *imports-resolved-classes* (make-hash-table :test 'equalp))
 
 (defun find-java-class (name)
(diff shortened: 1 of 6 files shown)


easye:
Awesome.  Need to integrate and test upstream.  Apologies.

easye:
Unfortunately there are two side effects with Alan's work:
From <http://abcl.org/trac/ticket/420#no3>:

1) With *INHIBIT-JSS-OPTIMIZATION* as nil, JSS won't compile any reference to a Java class that it not present in the classpath at compile time. Previously this wasn't the behavior

2) There are a slew of additional failures in the ANSI-TEST LAMBDA section, which indicates that we need to work through some wrinkles in the behavior of the patch.

LAMBDA.1, LAMBDA.2, LAMBDA.3, LAMBDA.4, LAMBDA.5, LAMBDA.6, LAMBDA.7,
LAMBDA.8, LAMBDA.9, LAMBDA.10, LAMBDA.21, LAMBDA.22, LAMBDA.54,
LAMBDA.57, LAMBDA.63, LAMBDA.64

easye:
Closed with <http://abcl.org/trac/ticket/420>.
<> rdfs:seeAlso <https://github.com/armedbear/abcl/commit/12f16229cd660d78a300a905c4383ced80577338> .

PR #15, opened 2017-01-01
some frames don't have file names, which was leading to a null pointe…

alanruttenberg:
…r exception when a string was made for the (null) pathname


alanruttenberg:
fixes #20

easye:
Merged with <https://github.com/armedbear/abcl/commit/55066d1e4c250295954baa432d77e927ecfdd116> <http://abcl.org/trac/changeset/14953>

PR #18, opened 2017-01-05
fixes #17

### LINKED ISSUE #3: getFunctionClassBytes for CompiledClosures <http://abcl.org/trac/ticket/414>
Can't seem to make it work. This is code that works for other functions

```
    (let ((classloader (#"getClassLoader" class)))
      (if (or (java:jinstance-of-p classloader "org.armedbear.lisp.MemoryClassLoader")
          (java:jinstance-of-p classloader "org.armedbear.lisp.FaslClassLoader"))
      (system::disassemble-class-bytes 
       (#"getFunctionClassBytes" classloader class))
```

### LINKED ISSUE #17: bug in invoke-restart <http://abcl.org/trac/ticket/427#ticket>
See the comment at 
https://github.com/sbcl/sbcl/blob/88fd8640e118ace44185a90baac095025acec3c5/src/code/target-error.lisp#L165

The issue is when invoke-restart is called with an actual restart with a test condition rather than a symbol. Currently invoke-restart calls (compute-restarts nil). So in the case where you've computed restarts (based on the condition) and then choose one of them, that restart can be considered inactive because the test function is run again with condition=nil instead of the condition that was signaled (and tested for).

alanruttenberg:
Define map-restarts from the body of compute-restarts, but with an extra argument to say whether to call the test function.

Redefine compute-restarts to use it (call-test-p t)

In invoke-restart and invoke-restart-interactively, if the argument is a restart-object just see if it is found without calling the test function. Otherwise do the usual.

### DIFF
--- a/src/org/armedbear/lisp/restart.lisp
+++ b/src/org/armedbear/lisp/restart.lisp
@@ -62,21 +62,26 @@
      ,@forms))
 
 (defun compute-restarts (&optional condition)
+  (let ((res ()))
+    (map-restarts (lambda(restart) (push restart res)) condition t)
+    (nreverse res)))
+
+(defun map-restarts (fn condition call-test-p)
   (let ((associated ())
 	(other ()))
     (dolist (alist *condition-restarts*)
       (if (eq (car alist) condition)
 	  (setq associated (cdr alist))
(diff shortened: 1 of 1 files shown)


alanruttenberg:
I mean, fixes #17

PR #23, opened 2017-01-05
Make compiling error message less cryptic

### LINKED ISSUE #22: Cryptic error while compiling <http://abcl.org/trac/ticket/431#ticket>
In slime, compiling region with, e.g.,  `(defun foo () (return-from bar nil))` gives a message about there being a program error with no further information

alanruttenberg:
fixes #22


easye:
Merged with <http://abcl.org/trac/changeset/14951> <https://github.com/armedbear/abcl/commit/a44ccb34b023416298333e56afb861a4eaf75e04>

PR #25, opened 2017-01-13
JSS syntax for fields

alanruttenberg:
<thing> is either {<lisp expression>} or a class name or abbreviation that find-java-class can use
  If <thing> is a lisp expression, then it is evaluated (in the lexical environment) and used as an instance
  If <thing> is a class name the result of find-java-class is used and a static field access is done.

<field> is either {<lisp expression} or string
  If <field> is a lisp expression it should evaluate to a string that names a field
  If <field> is a string (no quotes) it is used as the field name

eg. #"foo.bar.baz" -> (get-java-field (find-java-class 'foo.bar) "baz" t)
    #"{foo}.baz" -> (get-java-field (find-java-class foo) "baz" t)


alanruttenberg:
Would have done this to gitlab, but push to master is being rejected.

    13217 lori@den-hackintosh ~/repos/abcl/ % git push gitlab HEAD:master
    git push gitlab HEAD:master
    Counting objects: 8, done.
    Delta compression using up to 8 threads.
    Compressing objects: 100% (8/8), done.
    Writing objects: 100% (8/8), 2.48 KiB | 0 bytes/s, done.
    Total 8 (delta 4), reused 0 (delta 0)
    remote: GitLab: You are not allowed to push code to protected branches on this project.        
    To https://gitlab.common-lisp.net/abcl/abcl.git
     ! [remote rejected]   HEAD -> master (pre-receive hook declined)

easye:
Closed with <https://github.com/armedbear/abcl/commit/1028ea22b45510c950d98d6ad252ce0088a995a4>

Ferada:
The tests need some fixing, the last one should be using `JAVA:JSTATIC` since that's not imported; the penultimate references `CL-USER::PRINT-DB`, undefined and shouldn't be in that package anyway. Can it be removed or what's the intention? The timing loop also looks like it depends on the machine? At least on my laptop it's about ten times what the tests requires here.

Fixes here https://github.com/Ferada/abcl/commit/78414260a3ecba93673a8181e37274aa2bcaba31 but I'll make a separate PR anyway.

PR #13, opened 2016-12-14
Figured out how to let arglist get saved during build. Record source …

alanruttenberg:
…etc. for top-level closures. Also, once #'delete is available, don't allow duplicate source entries

PR #27, opened 2017-01-17
Use this pull instead of Pull #13 - make arglist get saved during build

alanruttenberg:
I rebased the remaining change on master.

### DIFF
--- a/src/org/armedbear/lisp/precompiler.lisp
+++ b/src/org/armedbear/lisp/precompiler.lisp
@@ -1191,29 +1191,35 @@
                            ,@decls
                            ,@(when doc `(,doc))
                            (block ,block-name ,@body))))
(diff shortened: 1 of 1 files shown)


easye:
As far as I can tell, the contents of this patch already exist in SVN, so rejecting.

easye:
After discovering in <https://github.com/armedbear/abcl/issues/29> that not all parts in SVN commits are necessarily present in <github.com/armedbear/abcl:master>, I need to re-examine this request with filesystem diffs.

alanruttenberg:
Yes, I compared to master before submitting.

easye:
Merged with <https://github.com/armedbear/abcl/commit/76d075ee23a6ab4a06a74a642127553a242f9b87>

PR #30, opened 2017-01-17
Fix regression of ansi test describe.x due to abcl-introspect function name computation

alanruttenberg:
ansi-test describe.* were failing because they tried to describe
'cl:lambda, and there was an error printing the symbol-function.
That's documented at http://abcl.org/trac/ticket/438

In the process of tracking this down I found other cases that might be
problematic and so revised any-function-name, maybe-jss-function, and
print-object to avoid those problems.


easye:
Merged with 
<http://abcl.org/trac/changeset/14958>
<https://github.com/armedbear/abcl/commit/6a5549c2c12492004915182452fa599d922b1279>

PR #32, opened 2017-01-20
stacktrace bugfix

alanruttenberg:
thought this was in, but not. Fixes a case where stacktrace uses a package name when there isn't a package.


easye:
Merged with https://github.com/armedbear/abcl/commit/79399eb10b49b43873f450373db39afd8edf3dcb

PR #33, opened 2017-01-21
printing macro objects fix

alanruttenberg:
printing of MacroObjects was broken (returned by, e.g., sys::environment-all-functions) . Also bit of customization for function printing. (setting *function-print-object-prefix* to "" should make print-object for functions agree with #"printObject" at least for top-level functions)


easye:
Merged with https://github.com/armedbear/abcl/commit/86926a5287a16df3e384113e8742a72732aa1f8a

PR #35, opened 2017-01-23
chain is documented to return the last value, but wasn't. Now it does

alanruttenberg:


easye:
Merged with <https://github.com/armedbear/abcl/commit/2d35b92b6d8d643695f2fe5afba1fd72b7d5ee44>

PR #36, opened 2017-01-23
a component :jdk-jar, a jar file where the pathname and name are rela…

alanruttenberg:
…tive to java home


alanruttenberg:
Couldn't figure out any other way to force asdf to look at a file outside the system directory.

easye:
Merged with <https://github.com/armedbear/abcl/commit/1faa514ac9a4bdf6a4cbe1d151bf4afc83be3d51>

PR #37, opened 2017-02-01
Fix a couple of places where user should be able to control things

alanruttenberg:
In quicklisp-abcl, quicklisp directory was defaulting to (user-homedir-pathname). Make this customizable by using variable asdf::*quicklisp-parent-dir*
in abcl-contrib.lisp there was defparameter *verbose*, which should be defvar, and which was ignored in two places where it shouldn't have been.

### DIFF
--- a/contrib/quicklisp/quicklisp-abcl.asd
+++ b/contrib/quicklisp/quicklisp-abcl.asd
@@ -8,9 +8,11 @@
     :version "0.4.0"
     :components nil)
 
+(defvar *quicklisp-parent-dir* (user-homedir-pathname))
+  
 (defmethod perform ((o load-op) (c (eql (find-system 'quicklisp-abcl))))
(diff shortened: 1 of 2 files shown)


easye:
Rejected as presented:

1) Should have been two separate commits, one for core ABCL and one for QUICKLISP-ABCL

2) Don't pollute ASDF package with ABCL-specific symbols (define and use a QUICKLISP-ABCL package instead).

3) Rename SYSTEM:*VERBOSE* to SYSTEM:*ABCL-CONTRIB-VERBOSE* to hopefully clarify its meaning.

Functionality split into two patches that have been merged

<https://github.com/armedbear/abcl/commit/5044bc70d699f69170ccefd64ee7e08be3bcb2ab>

<https://github.com/armedbear/abcl/commit/6cc94a54cf9256b2a0f13857d4c448d3b5c044fc>

Thanks for the patches!

PR #39, opened 2017-02-10
fix getParts for java array. 

alanruttenberg:
Was calling new simplestring on the index, instead of the string value of the index, creating index-long arrays full of nulls


easye:
Merged with <https://github.com/armedbear/abcl/commit/878a55d499f0d43c4ce41cff369c05df9054d46e>

PR #41, opened 2017-04-25
Signaling a simple error would signal another error if the message happened to have a ~ in it

alanruttenberg:
since in simple cases there are no format args, but format is called with the message string. I searched for all cases in the source where there were calls to setFormatControl(x) followed by setFormatArguments(NIL). In such cases I changed setFormatControl(x) -> setFormatControl(x.replaceAll("~","~~")). An example that would trigger the double error was compiling (defun foo (x) ("This should be quoted ~a" x)).

### DIFF
--- a/src/org/armedbear/lisp/ArithmeticError.java
+++ b/src/org/armedbear/lisp/ArithmeticError.java
@@ -72,7 +72,7 @@ public class ArithmeticError extends LispError
     public ArithmeticError(String message)
     {
         super(StandardClass.ARITHMETIC_ERROR);
-        setFormatControl(message);
+        setFormatControl(message.replaceAll("~","~~"));
         setFormatArguments(NIL);
         setOperation(NIL);
         setOperands(NIL);
--- a/src/org/armedbear/lisp/Condition.java
+++ b/src/org/armedbear/lisp/Condition.java
@@ -98,7 +98,7 @@ public class Condition extends StandardObject
   {
     super(StandardClass.CONDITION);
     Debug.assertTrue(slots.length == 2);
-    setFormatControl(message);
+    setFormatControl(message.replaceAll("~","~~"));
     setFormatArguments(NIL);
   }
 
--- a/src/org/armedbear/lisp/ControlError.java
+++ b/src/org/armedbear/lisp/ControlError.java
@@ -46,7 +46,7 @@ public final class ControlError extends LispError
     public ControlError(String message)
     {
         super(StandardClass.CONTROL_ERROR);
-        setFormatControl(message);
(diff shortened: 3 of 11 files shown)


easye:
While, I think I understand what sort of behavior you are trying to address here, I can't seem to see a "before and after" difference.

Could you please expand on the test case?

The test form as you wrote it
```
 (defun foo (x) ("This should be quoted ~a" x))
```
doesn't compile, but doesn't seem to give a "double error" either with or without this changeset applied.

I tried to guess at how to otherwise trigger the double error:

```
  (handler-case 
      (signal (make-condition 'simple-error :format-control "This should be quoted ~"))
    (t (e)
      (format nil "~a" e))))
```
but this seems to sensibly return 

```
error in format: String ended before directive was found.
  This should be quoted ~
                        ^
```

Can you help my understanding here by refining the test case to trigger the double error please?

alanruttenberg:
Try (funcall '("This should be quoted ~a" x))

alanruttenberg:
Put this:

    (defun foo (x) (funcall x))
into a file. Compile and load the file in slime.

    (funcall '~) 

I think that should work, or 
    
    (setf (symbol-plist '~) (list 1))(remprop '~ '~)

You can tell there will be problems just looking at the code. 
https://github.com/armedbear/abcl/blob/master/src/org/armedbear/lisp/Lisp.java#L2037
The ProgramError constructor uses the single argument method, which sets format args to nil. But embedded in the error string is the name of an arbitrary symbol, which could include "~".

The problem was noticed while debugging in slime. The error shown was  a format error. After some consternation I realized that it was a program-error that was the problem - I wasn't calling format in the code. I then looked at code for ProgramError, realized that it was buggy, and chose this as a solution. The alternative was tracking down any case that might embed a tilde and fixing it at the source, but that wouldn't protect against future transgression.

easye:
Replication recipe for posterity:  

Place the following in a file
```
(defun foo (x) (funcall x))
```
CL-USER> (compile-file "/Users/evenson/work/abcl/double-error") 
; Compiling /Users/evenson/work/abcl/double-error.lisp ...
; (DEFUN FOO ...)
; Wrote /Users/evenson/work/abcl/double-error.abcl (0.015 seconds)
#P"/Users/evenson/work/abcl/double-error.abcl"
NIL
NIL
CL-USER> (foo '~)
```
 gives
```
Error (FORMAT-ERROR) during printing: #<PROGRAM-ERROR {264F63C5}>
   [Condition of type PROGRAM-ERROR]
```

easye:
Merged with https://github.com/armedbear/abcl/commit/6b5d59d6b59f13b6efcf512997c244a4dc70451b

PR #42, opened 2017-05-14
More fixes.

Ferada:
Some more fixes. With this I can get the demo from CL-CFFI-GTK to run, even with some more complicated JNA/CFFI things.

The `SHARED-INITIALIZE` change is something that at least SBCL/CCL do this way, but I'm not sure this is the right way to fix it.


easye:
Merged.

PR #43, opened 2017-05-31
More patches.

Ferada:
Mixed bag again. I've not observed regressions with this. Tested on Linux and Java 1.8 only.


easye:
Merged with http://abcl.org/trac/changeset/15023 https://github.com/armedbear/abcl/commit/84a73de0e95e1fa544a407b956f3be113559b9a9 and following.

Thanks for the patches!

PR #44, opened 2017-06-02
More patches.

Ferada:
A couple more tickets and one FIXME.

For http://abcl.org/trac/ticket/233 - `DIRECTORY` just returns an empty list on SBCL, but the code already present would like to raise a `FILE-ERROR`, so I fixed it in that direction, but I'm not convinced that's the best option really.

### DIFF
--- a/src/org/armedbear/lisp/Lisp.java
+++ b/src/org/armedbear/lisp/Lisp.java
@@ -1868,10 +1868,11 @@ public final class Lisp
   {
     if (obj instanceof Package)
       return (Package) obj;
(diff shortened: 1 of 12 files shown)


easye:
Merged as <http://abcl.org/trac/changeset/15034> aka 
<https://github.com/Ferada/abcl/commit/4b6cbb1a917fbd33d949da131b57c93641fafbd6>
through <http://abcl.org/trac/changeset/15038> <https://github.com/armedbear/abcl/commit/d0cb69bb3a2497af8281fa591bae4ba50b539924> .

Thanks as always.

N.b. I might have somehow messed up a merge on one of your commits to the pathname test suite.

PR #46, opened 2017-06-07
Doc fixes

Ferada:
Three things:

- Updates the Makefile to make it the same usable sequence from the README file.
- Fixes one case of wrong rendering in LaTeX due to the `<` and `>` characters. [See this for an explanation](https://tex.stackexchange.com/questions/2369/why-do-the-less-than-symbol-and-the-greater-than-symbol-appear-wrong-as).
- Lastly regenerates the LaTeX sources with the former in place; the big removal there is actually correct - in `master` that block is duplicated!

Example output is http://macrolet.net/data/abcl.pdf - note that I also [installed the `cm-super` package](https://tex.stackexchange.com/questions/7185/how-to-make-the-pdf-output-of-better-print-quality/7188#7188), so perhaps a different font as explained there, `lmodern`, might be better in the long term.

### DIFF
--- a/doc/manual/Makefile
+++ b/doc/manual/Makefile
@@ -14,12 +14,16 @@ ABCL	= ../../abcl
 
 # XXX TODO 1) use more compact iterated form, 2) pare down to using --noinit
 grovel:
-	$(ABCL) --batch --noinform --load "grovel.lisp" \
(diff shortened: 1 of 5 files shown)


easye:
Merged with <http://abcl.org/trac/changeset/15044> and <http://abcl.org/trac/changeset/15045>.

PR #47, opened 2017-06-08
warn instead of format for flet removed warning

alanruttenberg:
Wasn't able to be muffled


easye:
Merged with https://github.com/armedbear/abcl/pull/47/commits/5e6d61cc4220ef426f19f3026b7ad2bbfa412c5b

PR #48, opened 2017-06-08
Java class specific print-object

alanruttenberg:
Specialize print-java-object-by-class to use java class-specific print method

(defmethod java::print-java-object-by-class ((class (eql ':|uk.ac.manchester.cs.owl.owlapi.concurrent.ConcurrentOWLOntologyImpl|)) obj stream)
	   (print 'hi)
         (call-next-method))

I don't know why there are 4 commits here. Only the last one is what I tried to get.


Ferada:
Is there a better way to do this, the loop there looks slow?

easye:
Merged via https://github.com/armedbear/abcl/pull/48/commits/437803a91890b8cfbeee1ba54efa05fbb5d22a1e

PR #52, opened 2017-06-08
#1".." reader for more complicated java expressions, using javaparser

alanruttenberg:
This is the base class for using the java parser library. JSS will use this (in another commit) to enable writing more complicated java expressions using #1"..." e.g.

(#1"new ByteBuddy().subclass(Object.class,t)
   .method(ElementMatchers.named("toString"))
   .intercept(FixedValue.value("Hello World!"))
   .make()
   .load(getClass().getClassLoader())
   .getLoaded()" )


easye:
I'll merge this work based on javaparser, but have doubts that this will work on Java 6 (but who knows). 

 If we have problems on non-Java 8, we will need to conditionalize its usage in JSS based on whether the javaparser jar can be loaded, as a lot of the other contribs depend on JSS.

easye:
I had to refactor the JAVAPARSER  and JSS systems to rationalize load behavior <https://github.com/armedbear/abcl/commit/64203af6ebc217b90285ab02fc44d7ea730e9bba>, otherwise loading JSS was failing for me.

In addition, you are missing the declarations for REPLACE-ALL and TREE-REPLACE in the javaparser code, so it doesn't work for other users at the moment.

alanruttenberg:
Ok. I will fix in my next patch. Unfortunately I can't do that until
Monday, I expect, as I'm out of town heading to an informal college reunion.

alanruttenberg:
Are you sure the dependency wasn't in the tests? I deliberately wrote this
so that it wouldn't be loaded unless used. Not so keen on redefining a
function in a second system .

easye:
Further fixes, including "stacking" the JAVAPARSER READ-INVOKE definitions in <https://github.com/armedbear/abcl/commit/80f8222a92cdc0ce03fd55757dd37b4d998f450d>

PR #54, opened 2017-06-15
restart for when methods are defined over functions or macros, and when arglists don't match

alanruttenberg:
One less annoyance


easye:
Merged with <https://github.com/armedbear/abcl/commit/58c69af7caa5ae0172fba3449777a221123a136f>

PR #56, opened 2017-06-15
Optimizations for jss field accessors

alanruttenberg:
Optimizations for jss field accessors, including "==" scope for one-only lookup of fields. New doc:

```
JSS syntax for fields
#"[<scope>]<thing>.<field>"

<scope> is empty or "==". scope is only paid attention to when <field> is a literal string

<thing> is either {<lisp expression>} or a class name or abbreviation that find-java-class can use
  If <thing> is a lisp expression, then it is evaluated (in the lexical environment) and used as an instance
    when <scope> is "==" you promise that instance will always be of the same class, and so field lookup
    is done once and cached. 
  If <thing> is a class name the result of find-java-class is used and a static field access is done.
    when <scope> is "==" you promise the static field is final and so the result is wrapped in (load-time-value  ...)

<field> is either {<lisp expression} or string
  If <field> is a lisp expression it should evaluate to a string that names a field
  If <field> is a string (no quotes) it is used as the field name

eg. #"foo.bar.baz" -> (get-java-field (find-java-class 'foo.bar) "baz" t)
    #"{foo}.baz" -> (get-java-field (find-java-class foo) "baz" t)
    #"==foo.baz" -> (load-time-value (get-java-field (find-java-class "foo") "bar" t))
    #"=={foo}.baz" -> TL;DR (only look up baz field once based on class of foo, and cache)
```

### DIFF
--- a/contrib/jss/invoke.lisp
+++ b/contrib/jss/invoke.lisp
@@ -208,7 +208,7 @@ NAME can either string or a symbol according to the usual JSS conventions."
     (unread-char char stream)
     (let ((name (read stream)))
       (if (or (find #\. name) (find #\{ name))
-          (jss-transform-to-field name)
+          (jss-transform-to-field name arg)
           (let ((object-var (gensym))
                 (args-var (gensym)))
             `(lambda (,object-var &rest ,args-var) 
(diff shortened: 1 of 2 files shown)


easye:
Merged with <https://github.com/armedbear/abcl/commit/9192422bc1f6254e173d6a830b0a15472e85bde2>.

PR #57, opened 2017-06-15
New macro with-class-lookup-disambiguated

alanruttenberg:
Suppose you have code that references class using the symbol 'object, and this is ambiguous. java.lang.Object or org.omg.CORBA.Object. Use 
  `  (with-class-lookup-disambiguated (lang.object) ...)
`
Within dynamic scope, find-java-class first sees if any of these match, and if so uses them to lookup the class.


easye:
Merged with <https://github.com/armedbear/abcl/commit/5713935299867b42a691d08450ac8d3cee5a7c1b> and <https://github.com/armedbear/abcl/commit/5713935299867b42a691d08450ac8d3cee5a7c1b>.

PR #61, opened 2017-07-12
define undefined conditions

alanruttenberg:
on malformed destructuring-bind, bogus-sublist-error was being signaled, said condition didn't exist, and so true error was masked.

On review this code seems to be adapted from CMUCL, but special casing error-fun = 'error and using program-error instead of the conditions. 

I brought over the conditions from CMUCL, which are in a comment at the bottom - they should loaded only after clos - wasn't sure of the mechanics. 

I then defined a function for bogus-sublist error, based on the reports for the condition, and replaced arg-count-error with the more detailed report as well.

### DIFF
--- a/src/org/armedbear/lisp/destructuring-bind.lisp
+++ b/src/org/armedbear/lisp/destructuring-bind.lisp
@@ -69,6 +69,25 @@
          :format-control "Wrong number of arguments for ~S."
          :format-arguments (list name)))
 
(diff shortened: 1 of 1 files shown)


easye:
Merged with <https://github.com/armedbear/abcl/commit/38884a0a1fe7170c7721822bb4ad2170fb49806e>.

PR #65, opened 2017-09-17
read-sharp-java-expression bugfixes

alanruttenberg:
bugfixes: 1) method call expression lookup java class for jstatic. 2_ maybe-class, if it isn't a class, intern in current *package* vs. jss

### DIFF
--- a/contrib/jss/jss.asd
+++ b/contrib/jss/jss.asd
@@ -11,6 +11,7 @@
 				     (:file "optimize-java-call")
                                      (:file "classpath")
 				     (:file "transform-to-field")
-                                     (:file "compat"))))
+                                     (:file "compat")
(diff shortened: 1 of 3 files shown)


Ferada:
How do you actually test this when `replace-all` is completely undefined when loading `javaparser`?

alanruttenberg:
@ferada, indeed. Was in some other code in my fork. Defined the missing functions in jss.

Ferada:
@alanruttenberg okay cool, is that in `util.lisp`? I think that's still not added to this branch yet?

easye:
I have now merged in 8cc9746 adding `util.lisp` and normalized whitespace to a good version for further comparison with github.com/alanruttenberg/abcl

Issue #66, opened 2017-09-20
Bug in logic of asdf :mvn resolution


alanruttenberg:
If you don't explicitly provide a repository as part of the :mvn specification in a sysdef, resolve-dependencies is called (from resolve) with repositories nil.

Then, in resolve-dependencies, since repositories-p is t and repositories is nil, the collect-request doesn't get a repository added to it. Subsequently, if the dependency isn't already downloaded, the resolution fails.

I don't understand the intended behavior if repositories-p is t and repositories is nil. I don't see anything documented about this. If repositories is nil, I think the default repository should be used. Failing that, resolve shouldn't pass the repositories argument, if nil.

That I wasn't adding the repository to the collect request is why my mvn-module was failing in the same way. Patch for that forthcoming.

alanruttenberg:
Also, it looks like, at least before the offered patch, that *mvn-default-repository* was being set to nil somewhere, even after (ensure-remote-repository) was setting it.

PR #67, opened 2017-09-20
collect request needs a repository

alanruttenberg:
See https://github.com/armedbear/abcl/issues/66 - this patches abcl-asdf:resolve-multiple-dependencies to add a repository to the collect request, and patches abcl-asdf:resolves to not call resolve-dependencies with :repositories nil if repositories is nil


easye:
Closed with https://github.com/armedbear/abcl/commit/b200a22a452264c18253596dea4e6db48a8853ea

PR #73, opened 2017-10-16
Add jtypecase form, fix print-object of null pointer

alanruttenberg:
Like typecase except the check is whether the form is an instance of one of the classes, which are lookup up using find-java-class. 

Also included: a minor fix to print-java-object-by-class - it was getting an error while printing a null pointer.

### DIFF
--- a/contrib/jss/jss.asd
+++ b/contrib/jss/jss.asd
@@ -11,6 +11,7 @@
 				     (:file "optimize-java-call")
                                      (:file "classpath")
 				     (:file "transform-to-field")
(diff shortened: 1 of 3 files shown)


Ferada:
IMO nice, though I'd personally say `#<null>` might be preferable instead of introducing a new word here.

alanruttenberg:
Agreed, suggestion taken.

easye:
Merged with https://github.com/armedbear/abcl/commit/8b1b395e97a8210375a4c11c6ab0378cb9c7f7a1

PR #76, opened 2017-11-08
Fix translate-directory-components-aux: throw takes 2 arguments

metawilm:


easye:
Nice to have a more concrete reference to the test/issue this fixes, but I will have a look at this soon.

Thanks for the patch!

easye:
Closed via https://github.com/armedbear/abcl/commit/c1cd44705fc5a3df934fac9687414675d05e2407

Thanks again for the patch.

PR #77, opened 2017-11-14
BUGFIX: invoke-restargs compiled was evaluating the first argument twice :-(.

alanruttenberg:


alanruttenberg:
Example to show bug:

(defun foo () (#"replaceAll" (progn (print 'hi) "foo") "foo" "bar"))

Before fix, hi printed twice, after once.

easye:
Merged with https://github.com/armedbear/abcl/commit/37fb7e6c94cb3b5e0ea0cd0d05c18c76ec195362

Issue #85, opened 2018-04-22
READER-ERROR - FORMAT-CONTROL is preformatted


phoe:
ABCL 1.5.0

```common-lisp
CL-USER> (handler-case (read-from-string "cl:asdfsad") (error (e) (describe e)))

#<READER-ERROR {B20D3DF}> is an instance of #<STANDARD-CLASS READER-ERROR {B0459D1}>.
The following slots have :INSTANCE allocation:
  FORMAT-CONTROL     "The symbol \"ASDFSAD\" was not found in package CL."
  FORMAT-ARGUMENTS   NIL
  STREAM             #S(SYSTEM::STRING-INPUT-STREAM)
```
The `FORMAT-CONTROL` string is preformatted, and it should not be.

`FORMAT-CONTROL` should be `"The symbol ~S was not found in package CL."` and `FORMAT-ARGUMENTS` should be `("ASDFSAD" CL)`.

PuercoPop:
Digging a little bit the error is in [Stream.java](https://github.com/armedbear/abcl/blob/master/src/org/armedbear/lisp/Stream.java#L1194-L1202). If the symbol was not external to a package you should get a similar error.

However the class [ReaderError](https://github.com/armedbear/abcl/blob/master/src/org/armedbear/lisp/ReaderError.java) doesn't appear to have a constructor that allows one to pass format arguments. Should we just create another constructor that takes 'rest' arguments as the last argument? Or would it be prefereable to create a different constructor for the one argument case only?

phoe:
I don't think I'm the proper person to answer your questions. I'm not competent in ABCL internals.

easye:
Merged in <https://github.com/armedbear/abcl/commit/2e80b56d7384dd42e03d580cdd79872c9af13d32>

There is no consistency in adding constructors for the Java-side implementations of things.  We essentially just add constructors as needed.

PR #86, opened 2018-04-23
ReaderError: Don't evaluate format control string when signaling the error

PuercoPop:
This PR is taking a stab at fixing #85.

What I've done is add a new constructor to Reader Error with a new constructor that takes two extra arguments  that are passed to the format control string. This is not an ideal solution as I'm only handling a two arguments for the format control string case. Not even checking the number of arguments needed by the format control string.

Another thing I'm wondering is whether we should we look up the package so we can referrer to it by its name. With the code as is submitted if the package was being referred to by a nickname that is how it is going to show up in the ReaderError Condition. As data points, SBCL uses the string "COMMON-LISP" and CCL uses the package object itself.

### DIFF
--- a/src/org/armedbear/lisp/ReaderError.java
+++ b/src/org/armedbear/lisp/ReaderError.java
@@ -52,6 +52,14 @@ public final class ReaderError extends StreamError
         setStream(stream);
     }
 
+  public ReaderError(String message, Stream stream, LispObject arg1, LispObject arg2)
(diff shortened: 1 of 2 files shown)


easye:
Merged in <https://github.com/armedbear/abcl/commit/2e80b56d7384dd42e03d580cdd79872c9af13d32>

Thanks for the patch!

PR #89, opened 2018-07-27
stackframe head edge case fix

alanruttenberg:
Fixes a case where instead of seeing your error in the stacktrace you see an error introduced by this bug.


easye:
Closed with <https://github.com/armedbear/abcl/commit/915c45c3d89e2a6bedf7e95d144a113e5d06c5e8>

PR #103, opened 2019-10-11
Signal better error for out-of-bounds `REPLACE`.

Ferada:
Simple and not particularly pretty fix for https://abcl.org/trac/ticket/348. Better than asserting an potentially invalid type though.

### DIFF
--- a/src/org/armedbear/lisp/replace.lisp
+++ b/src/org/armedbear/lisp/replace.lisp
@@ -156,9 +156,13 @@
                                 ((:end2 source-end)))
   "The target sequence is destructively modified by copying successive
 elements into it from the source sequence."
+  (check-type target-start (integer 0 #.array-total-size-limit))
+  (check-type source-start (integer 0 #.array-total-size-limit))
+  (check-type target-end (or null (integer 0 #.array-total-size-limit)))
+  (check-type source-end (or null (integer 0 #.array-total-size-limit)))
   (let ((target-end (or target-end (length target-sequence)))
 	(source-end (or source-end (length source-sequence))))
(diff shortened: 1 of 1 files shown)


src/org/armedbear/lisp/replace.lisp
```diff
@@ -156,6 +156,10 @@
                                 ((:end2 source-end)))
   "The target sequence is destructively modified by copying successive
 elements into it from the source sequence."
+  (check-type target-start (integer 0 #.most-positive-fixnum))
```
PuercoPop: Afaik indices don't have an upper limit so we shouldn't have the most-postive-fixnum as an upper bound. The type specifier `(integer 0 *)` would be achive the goal of asserting the type of the index to be a non-negative integer.

Ferada: I'm afraid that particular assumption is present in _a lot_ of other places, even in this same file. I don't mind changing it to unbounded, but not so much making that same change for all other instances.

PuercoPop: I spoke too soon, there is indeed an upper limit [ARRAY-TOTAL-SIZE-LIMIT](http://www.lispworks.com/documentation/HyperSpec/Body/v_ar_tot.htm) and it is defined in [arrays.lisp](https://github.com/armedbear/abcl/blob/master/src/org/armedbear/lisp/arrays.lisp#L34) to be the most-positive-fixnum. Consider `(integer 0 #.array-total-size-limit)` to keep things DRY.

easye:
Merged.  Thank you very much.

PR #104, opened 2019-10-11
Check for element type before filling vectors.

Ferada:
Attempting to fix https://abcl.org/trac/ticket/297.

```
(make-array 5 :element-type '(unsigned-byte 8) :initial-element 300)
=>
  The value 300 is not of type (UNSIGNED-BYTE 8).
;; was #(44 44 44 44 44)

(make-array 5 :element-type '(unsigned-byte 16) :initial-element (+ 300 (expt 2 16)))
=>
  The value 65836 is not of type (UNSIGNED-BYTE 16).
;; #(65836 65836 65836 65836 65836)
```


easye:
Merged. Thanks!

PR #110, opened 2019-11-21
Easye/test format dollar

easye:


easye:
Test of reversing the direction of the `svn` bridge.

PR #113, opened 2019-11-21
Incomplete work on the CI using AdoptJDK

easye:
Downloading the AdoptJDK image using via `jenv` in a 2x2 build matrix.


easye:
Still not invoking the correct openjdk via jenv, but at least it has the 2x2 matrix working

PR #122, opened 2019-11-24
Fix setting JDK directory for non-Travis situations

easye:


easye:
Reviewed.  

TODO: meaningful ASDF signaling to report test failures accurately

Issue #95, opened 2019-02-11
DECODE-FLOAT returns single float significand for double float argument


robert-dodier:
CLHS says "decode-float computes three values that characterize float. The first value is of the same type as float and represents the significand."

I find that DECODE-FLOAT applied to double float arguments always returns a single float significand instead of a double float.

```
CL-USER(5): (type-of 1d0)        
DOUBLE-FLOAT
CL-USER(6): (type-of (decode-float 1d0))
SINGLE-FLOAT
CL-USER(7): (type-of most-positive-double-float)
DOUBLE-FLOAT
CL-USER(8): (type-of (decode-float most-positive-double-float))
SINGLE-FLOAT
CL-USER(9): (type-of least-positive-double-float)
DOUBLE-FLOAT
CL-USER(10): (type-of (decode-float least-positive-double-float))
SINGLE-FLOAT
```
I am working with ABCL built from svn as of r15125 if I am not mistaken. `lisp-implementation-version` reports: 
```
"1.6.0-dev"
"Java_HotSpot(TM)_Server_VM-Oracle_Corporation-1.8.0_60-b27"
"i386-Linux-3.13.0-44-generic"
```
```
CL-USER(6): *features*
(:X86 :JAVA-1.8 :UNIX :LINUX :ARMEDBEAR :ABCL :COMMON-LISP :ANSI-CL :CDR6 :MOP :PACKAGE-LOCAL-NICKNAMES)
```
`uname -a` reports: `Linux freekbox 3.13.0-44-generic #73-Ubuntu SMP Tue Dec 16 00:23:46 UTC 2014 i686 i686 i686 GNU/Linux`

robert-dodier:
Here is a patch to fix this problem and also #93.

[decode-float-single-double-patch.txt](https://github.com/armedbear/abcl/files/2849949/decode-float-single-double-patch.txt)

easye:
@robert-dodier Thanks for the diagnosis and the patch.  I must confess that I need to read up on the `decode-float` interface to take a stab at what is going on with the de-normalized float problems you identify in #94.

Issue #94, opened 2019-02-10
DECODE-FLOAT returns significand less than 0.5 for denormalized double floats


robert-dodier:
Related to #93 but it seems likely this is a different bit of code so I'm opening a separate report.

CLHS says, "decode-float divides float by an integral power of b so as to bring its value between 1/b (inclusive) and 1 (exclusive), and returns the quotient as the first value." b is the base of the floating point number system, so in the case of ABCL it is 2, therefore the signficand returned by DECODE-FLOAT should be greater than or equal to 0.5 and less than 1.

I find that for double floats less than LEAST-POSITIVE-NORMALIZED-DOUBLE-FLOAT divided by 2, DECODE-FLOAT returns significand less than 0.5.

```
$ ./abcl
Armed Bear Common Lisp 1.6.0-dev
Java 1.8.0_60 Oracle Corporation
Java HotSpot(TM) Server VM
Low-level initialization completed in 0.414 seconds.
Startup completed in 2.535 seconds.
Type ":help" for a list of available commands.
CL-USER(1): (lisp-implementation-version)
"1.6.0-dev"
"Java_HotSpot(TM)_Server_VM-Oracle_Corporation-1.8.0_60-b27"
"i386-Linux-3.13.0-44-generic"
CL-USER(2): (lisp-implementation-type)
"Armed Bear Common Lisp"
CL-USER(3): 
CL-USER(3): (decode-float least-positive-normalized-double-float)
0.5
-1021
1.0
CL-USER(4): (decode-float (/ least-positive-normalized-double-float 2))
0.5
-1022
1.0
CL-USER(5): (decode-float (/ least-positive-normalized-double-float 1024))
9.765625E-4
-1022
1.0
CL-USER(6): (decode-float (/ least-positive-normalized-double-float 4))
0.25
-1022
1.0
CL-USER(7): (decode-float (/ least-positive-normalized-double-float 8))
0.125
-1022
1.0
CL-USER(8): (decode-float (/ least-positive-normalized-double-float 16))
0.0625
-1022
1.0
CL-USER(9): (decode-float least-positive-double-float)              
2.220446E-16
-1022
1.0
```

I am working with ABCL built from svn as of r15125 if I am not mistaken.

```
CL-USER(6): *features*
(:X86 :JAVA-1.8 :UNIX :LINUX :ARMEDBEAR :ABCL :COMMON-LISP :ANSI-CL :CDR6 :MOP :PACKAGE-LOCAL-NICKNAMES)
[...]

Issue #93, opened 2019-02-10
DECODE-FLOAT returns significand less than 0.5 for all single floats


robert-dodier:
CLHS says, "decode-float divides float by an integral power of b so as to bring its value between 1/b (inclusive) and 1 (exclusive), and returns the quotient as the first value." b is the base of the floating point number system, so in the case of ABCL it is 2, therefore the signficand returned by DECODE-FLOAT should be greater than or equal to 0.5 and less than 1.

For all single floats that I tried, ABCL's DECODE-FLOAT returns significand less than 0.5. The value returned is on the order of 1e-9 for normalized single floats and smaller for denormalized single floats.

I am working with ABCL built from svn as of r15125 if I am not mistaken.

```
$ ./abcl
Armed Bear Common Lisp 1.6.0-dev
Java 1.8.0_60 Oracle Corporation
Java HotSpot(TM) Server VM
Low-level initialization completed in 0.437 seconds.
Startup completed in 2.462 seconds.
Type ":help" for a list of available commands.
CL-USER(1): (decode-float 1f0)
9.313226E-10
30
1.0
CL-USER(2): (decode-float 1f6) 
1.7763568E-9
49
1.0
CL-USER(3): (decode-float 1f-6)
9.765625E-10
10
1.0
CL-USER(4): (decode-float least-positive-normalized-single-float)
9.313226E-10
-96
1.0
CL-USER(5): (decode-float least-positive-single-float)
2.220446E-16
-97
1.0
```

`*FEATURES*`:
```
CL-USER(6): *features*
(:X86 :JAVA-1.8 :UNIX :LINUX :ARMEDBEAR :ABCL :COMMON-LISP :ANSI-CL :CDR6 :MOP :PACKAGE-LOCAL-NICKNAMES)
```

`uname -a` reports: `Linux freekbox 3.13.0-44-generic #73-Ubuntu SMP Tue Dec 16 00:23:46 UTC 2014 i686 i686 i686 GNU/Linux`

PR #129, opened 2019-11-29
[IN PROGRESS] Fix the numeric tower for floats

easye:
(Robert Dodier)

Address problems with DECODE-FLOAT

Includes all known patches from Robert Dodier.

Robert Dodier

References
<https://github.com/armedbear/abcl/issues/93>
<https://github.com/armedbear/abcl/issues/94>
<https://github.com/armedbear/abcl/issues/95>

`rdfs:seeAlso` <https://gitlab.common-lisp.net/abcl/abcl/blob/master/src/org/armedbear/lisp/numbers.lisp>.

### DIFF
--- a/src/org/armedbear/lisp/numbers.lisp
+++ b/src/org/armedbear/lisp/numbers.lisp
@@ -153,11 +153,34 @@
              :format-arguments (list float))))
 
 (defun decode-float (float)
-  (multiple-value-bind (significand exponent sign)
-      (integer-decode-float float)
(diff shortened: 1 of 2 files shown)


easye:
Application of all know patches to actually address the problem. 

On exploration I suspect there is something much deeper here, but can't prove it yet.

@slyrus

easye:
The results of PROVE tests for this problem <https://travis-ci.org/easye/abcl/jobs/618904711#L9648>

PR #136, opened 2019-12-17
Don't ever pass :repository nil and :repositories nil to resolve

alanruttenberg:


alanruttenberg:
I think I will withdraw this. As I recall I was getting an inscrutable error in certain conditions, probably in a situation in which there was a bug in setup.
If I try the below, resolution succeeds. The artifact is in my file system ~/.m2. I wasn't sure whether it should succeed in that case.

Otherwise, the error I get now is reasonable.
 
```lisp
(abcl-asdf::resolve-dependencies "net.sourceforge.owlapi" "owlapi-distribution"
                              :version "4.2.6"
                              :repository NIL
                              :repositories NIL)
```

easye:
Withdrawn by @alanruttenberg until we sketch out semantics.  

We have never fully specified what a generalized boolean `nil` should signal that the implementation should do for `:repository` or `:respositories` keywords.

PR #137, opened 2019-12-17
Add some things to gitignores

alanruttenberg:


easye:
We don't currently have a `test` subdirectory, and I can think of no reason to add one for everyone at this point.  

Please remove the `test/.gitignore` file from this patch.

PR #148, opened 2020-03-09
Easye/2090309b

easye:


easye:
Clearing the queue for abcl-1.6.1

Highlights

1. `jstatic` now narrows calls to method taking shorts, bytes, and longs well, but the same needs to be added to the `jcall` machinery
2.  abcl-asdf is now working if `abcl-asdf:*mvn-libs-directory*` is set on invocation, and working again on Java6

PR #153, opened 2020-03-18
Fix printing inherited package symbols; collect various tests

easye:


easye:
@alanruttenberg  This patch needs further work:  it causes a slew of additional failures in the ANSI-TESTS, and seems to bork `quickload` forms.  

I need to go over the CLHS to better understand the rules for omitting package prefixes when pushing symbols through the Lisp printer.  Basic stuff, but I find my understanding of what the rules should be a little lacking.  References to relevant CLHS sections welcome…

<https://travis-ci.org/github/easye/abcl/jobs/664507250#L17377>

easye:
Closing until someone can address the problems with the symbol patch that cause ~50 addtional failures in the ANSI-TESTS and borks `quickload`.

PR #170, opened 2020-04-24
build: support abcl.properties creation openjdk{6,7,8,11,13,14}

easye:
Additionally add test for openjdk14 to the Travis-CI build.

Improve script for creating build properties, which may be invoked
like:

    bash ci/create-abcl-properties.bash openjdk8

Currently this script has baked-in "knowledge" of options for various
platforms for now, but will eventually directly transcribe from the
"facts" in <file:abcl.rdf>.


easye:
part of #172

PR #174, opened 2020-05-06
build: can now use Maven to build

easye:
The basic assembly of <file:dist/abcl.jar> now works, although these
aren't completely hooked up to Maven's notion of output artifacts.

TODO:

We aren't really intercepting maven-jar-plugin default-jar goal
correctly, so the resulting <file:build/abcl-1.6.x-SNAPSHOT.jar>
artifact doesn't contain all artifacts.  We probably need to correctly
define our resources to Maven, but they are currently expressed as
various Ant properties, so we need to slave the Maven goals to
appropiate Ant tasks?

Maven really wants to create all output under a single directory,
while the Ant recipe puts intermediate artifacts under <file:build/>
and final results in <file:dist/>.

Need to somehow encapsulate the existence of
<file:dist/abcl-contrib.jar> within the <file:contrib/pom.xml> Maven
description, even though its build is part of the Ant `abcl.jar` task.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project
    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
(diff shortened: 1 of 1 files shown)


easye:
TODO need to update versions, but that will happen in the 1.9.0 metadata anyways.

PR #176, opened 2020-05-07
abcl-introspect: start collecting various dissassemblers

easye:
TODO: ensure that CL:DISSASSEMBLER can do do something.  If we can
load ABCL-INTROSPECT we should interogate the assembler hooks until we
work.

TODO: deprecate JAD dissassembler

Start collecting the download URIs for more common JAD archiectures.

TODO: genralize the introspect of the correct architecture at runtime.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ install:
 # TODO: figure out how to add abcl to our path
 
 script:
+  - ${ABCL_ROOT}/abcl --batch --load ${ABCL_ROOT}/ci/test-abcl-introspect.lisp
   - ${ABCL_ROOT}/abcl --batch --load ${ABCL_ROOT}/ci/test-cffi.lisp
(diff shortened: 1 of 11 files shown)


alanruttenberg:
I made a collection at some point. Thought I had code to exercise them but I can't find it.
The ones I found were: fernflower, cfr, and procyon. I have jad-1.5.8g-macosx as well. I expect they are easy to find on the web but can send you jars if not.

easye:
I am particularly interested if you have recipes for installing and/or installing the given tool as `cl:disassemble`.  

My plan is to gather some of the more common ones in `abcl-introspect`, and then provide some kind of introspection strategy for `cl:disassemble` so that it works with at least one.  If people are running from a JDK (which is pretty common these days I think, especially for a developer) we should probably default to `javap` rather than `jad`.

easye:
Working on collecting disassemblers now collected in #180 which collapses this work into a single patch

PR #203, opened 2020-05-27
The upper limit for 32 bit values is simply wrong.

easye:
This should signal an error:

    (make-array 16 :element-type '(unsigned-byte 32) :initial-element  (expt 2 32))


alanruttenberg:
Left hand / right hand.
```lisp
(let* ((type '(unsigned-byte 32))
       (array (make-array 16 :element-type type :initial-element  (expt 2 32))))
   (typep (aref array 0) type))
-> nil
```
Might be a good test for safety=3: Is the type of the first element of the array the same as the declared type?

PR #217, opened 2020-06-01
Easye/nio/20200601a

easye:
First alpha of the nio interface using just `cl:make-array` keyword arguments


easye:
supersedes #214

easye:
superseded by #219

PR #239, opened 2020-06-17
Add more boxing support for runtime class.

easye:
Supersedes #107 with the current test coverage.


easye:
TODO: gather the tests for ```JNEW-RUNTIME-CLASS``` to find the constructors we would typically use to best gauge how to use these new compilations.

PR #248, opened 2020-07-09
t/byte-vectors: test for failures found with NIBBLES

easye:


easye:
Superseded with fixes in #249

PR #258, opened 2020-07-27
Easye/t/20200722b

easye:
supersedes #257


easye:
Unfortunately, this doesn't really test anything for <https://abcl.org/trac/ticket/470>, as we would need to test that code compiled with `openjdk8` works on `openjdk11` and vice versa.  

We could potentially install multiple JVMs in each CI instance, then try to load the compiled implementation.  But this isn't that patch.

PR #265, opened 2020-08-06
Feature: serialization of functions

alessiostalla:
This implements serialization for functions:
- top-level and local
- interpreted and compiled
- closures included (provided closed-over objects are themselves serializable, of course)

Notable points worth a second look: I changed how loadedFrom is computed for functions loaded from FASLs, recording the fasl truename if present. Using *load-truename* only would break reading function class bytes for serialization (but also for disassembly).

A few (pretty minimal) tests are provided in Java and in Lisp.

What's missing:
- Test serialization/deserialization across ABCL images
  - Inlcuding re-serializing a deserialized function on a different image (could depend from a FASL that doesn't exist on the receiving image/machine, we should simulate that in a test)
- Test serialization/deserialization of mutually recursive local functions

Additional notes: getClassBytes for (local) functions loaded from a FASL depends on such FASL being still present when the function is disassembled or serialized. Top-level functions instead always carry the bytes with them. Maybe we should treat them uniformly according to a flag: *function-bytecode-strategy* which could be :load-time, :lazy or :disabled/nil. This could be used to reduce the footprint of functions that you don't want to disassemble or serialize, or to pre-load the bytes (saving disk reads) if you intend to serialize a function from the start.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,5 +4,8 @@ abcl
 abcl-test*log
 ext
 scratch
+.idea/
+*.iml
+target/
 \#*\#
 \.\#*
--- a/abcl.asd
+++ b/abcl.asd
@@ -57,7 +57,9 @@
                          #+abcl
                          (:file "runtime-class")
                          #+abcl
-                         (:file "package-local-nicknames-tests")))))
+                         (:file "package-local-nicknames-tests")
+                         #+abcl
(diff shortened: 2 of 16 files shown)


easye:
Looks good.  Thanks @alessiostalla !

PR #270, opened 2020-08-13
abcl-introspect: EXT:STREAM-UNIX-FD returns integer file descriptor

easye:
The integer can be passed meaningfully by passed to syscalls like
fdopen().

TODO: implement in the core application, current strategy needs JSS to
      reliably finad access private member fields


easye:
Passed CI in <https://travis-ci.org/github/easye/abcl/builds/717617035>

PR #283, opened 2020-09-16
Allow for java.version that reports an integer

easye:
Needed for running openjdk15 GA release <https://jdk.java.net/15/>.

Use configuration options for openjdk14.  

Subjectively seems rather zippier.


easye:
Superseded by #284

PR #284, opened 2020-09-17
Patches and CI for openjdk15

easye:
Use openjdk15 instead of openjdk14 under Travis CI.

Allow for java.version that reports an integer without a patchlevel
which is needed for running openjdk15.


easye:
Superseded by #285

PR #286, opened 2020-09-19
Easye/1.8.0/dev/20200919a

easye:
**** ABCL 1.8.0
***** TODO Current PATHNAME overhaul (tests)
(possibly with incomplete nested PathnameJar entries)

Probably easier to do efficiently with openjdk8: doesn't mean we can't
have an inefficient one.  In the current state it is a complete cache,
which if referenced by a weak reference, would be sufficient to use as
a fall-back.


easye:
Superseded by <https://github.com/armedbear/abcl/pull/289>.

PR #299, opened 2020-09-25
abcl-prove/t: ignore-errors clause now encloses test invocation

easye:
Add TIME forms to ABCL/TEST/ANSI invocation.

PR #300, opened 2020-09-26
ci: various maintenance changes

easye:
Travis CI now uses versions of STATIC-VECTORS, CL+SSL, and CFFI from
Quicklisp distribution as all outstanding patches have been absorbed
upstream.  While this means testing will change depending on the
current Quicklisp distribution when the tests are invoked, it will
more properly reflect common usage of these systems.

Overhaul ABCL-PROVE system definition explicitly enumerating tests for
more deterministic behavior.  Remove unnecessary ABCL-PROVE/T
sub-system.  Remove testing of JEANNIE from ABCL-PROVE so that it has
no dependencies external to the ABCL source tree.

Add TIME forms to ABCL/TEST/ANSI invocation.


easye:
Supersedes #299

PR #308, opened 2020-10-07
Ensure that SYS:CHECK-REDEFINITION cannot signal error conditions

easye:
Unfortunately, the ABCL SLIME v2.26 code can throw errors when compiling source in Emacs
buffers.  As currently used in SLIME, the pathname may have a device
containing the string "emacs-buffer" whose PATHNAME name denotes the
buffer, which often contain wild-pathname characters
(e.g. "*slime-scratch*").

Use IGNORE-ERRORS to be extra careful.


easye:
Superseded by #309

PR #309, opened 2020-10-08
Ensure that SYS:CHECK-REDEFINITION does not signal error conditions

easye:
Unfortunately, SLIME v2.26 code can throw errors when compiling Emacs
buffers, namely the pathname may have a device containing the string
"emacs-buffer" whose PATHNAME name denotes the buffer which often
contain wild-pathname characters (e.g. "*slime-scratch*") causing a
naieve PROBE-FILE to throw errors.

TODO redo ABCL SLIME's use of PATHNAME-DEVICE components to mark a
source location.

Use IGNORE-ERRORS to be extra careful.


easye:
Supersedes #308

PR #312, opened 2020-10-09
Abcl/1.8.0/rc/20201009/b

easye:
Operations on nested jars for locally available archives working for `probe-file`, `truename`, `file-write-date`, and `open`. Yay!

TODO winnow down the abcl/lisp/test suite to show this progress

TODO figuring out what disabling cache might me

Supersedes #311

PR #313, opened 2020-10-10
Abcl/1.8.0/rc/20201010/a

easye:
Supersedes #312


easye:
abcl-1.6.0 RC0 without metadate

PR #327, opened 2020-10-20
Implementation of accessing recursive entries in local jars

easye:
Now seemingly broken for accessing any content in remote jars.

SYS:CLEAR-ZIP-CACHE will clear the cache


easye:
Superseded by #328

Issue #345, opened 2020-11-03
ABCL 1.8.0 bug in vector-push-extend


brown:
The transcript below demonstrates the bug.  The output of the last print should be
`#(12345 23456 4294967295 2147483647)`.

```
chuwi [96028]  cat bug.lisp

(defun bug ()
  (let ((foo (make-array 0 :element-type '(unsigned-byte 32) :fill-pointer 0 :adjustable t)))
    (vector-push-extend 12345 foo)
    (print foo)
    (vector-push-extend 23456 foo)
    (print foo)
    (vector-push-extend #xffffffff foo)
    (print foo)
    (vector-push-extend (1- (ash 1 31)) foo)
    (print foo)))

chuwi [96029]  abcl --noinit
Armed Bear Common Lisp 1.8.0
Java 13.0.4 Private Build
OpenJDK 64-Bit Server VM
Low-level initialization completed in 0.654 seconds.
Startup completed in 3.507 seconds.
Type ":help" for a list of available commands.
CL-USER(1): (load "bug.lisp")
T
CL-USER(2): (bug)

#(12345) 
#(12345 23456) 
#(12345 23456 4294967295) 
#(23456 4294967295 0 2147483647) 
#(23456 4294967295 0 2147483647)
CL-USER(3): 
```

easye:
<https://github.com/easye/abcl/commit/a0011b62647d7883fab465112c8d79184e1c7cdb> fixes the immediate bug, but I would like to hunt down where exactly the underlying buffer position is being advanced from zero.

brown:
Thanks very much for fixing this bug!  The unit tests for my github projects all pass now with ABCL.

PR #346, opened 2020-11-04
Fix VECTOR-PUSH-EXTEND for unsigned-byte specializations

easye:
Addresses <https://github.com/armedbear/abcl/issues/345>.

Somehow the position marker on our specialized buffers is getting
"advanced" from zero under some condition that I can't immediately
find from looking through the code.

TODO figure out what is advancing the position, reset the underlying
buffer position to zero.


easye:
Superseded by #348

PR #349, opened 2020-11-04
arrays: fix fail to reset position after relative access

easye:
The position of all buffers should always be zero.  TODO how to
unintrusively make this assertion?

We fix new types specializing (or (unsigned-byte 8) (unsigned-byte
32)) in BasicVector_ByteBuffer, ComplexArray_ByteBuffer,
ComplexVector_ByteBuffer, and ComplexVector_IntBuffer for relative
accesses which always advance the position of the
underlying java.nio.Buffer subclass.

Fixes <https://github.com/armedbear/abcl/issues/345>.

--

Supersedes #345


easye:
Superseded by <https://github.com/armedbear/abcl/pull/350>

PR #352, opened 2020-11-10
Easye/fasl loading/20201110a

easye:
Starts to address situations where symbols in the fasl loader aren't in the current package when `cl:load` is invoked.  

Fails to fix loading a fasl when the current package doesn't `(use :cl)`


easye:
Superseded by #353.

PR #353, opened 2020-11-11
Easye/fasl loading/20201111a

easye:
This seems to work to fix our fasl loading problems 

Supersedes #352.


easye:
Superseded by #354

PR #361, opened 2020-12-21
Support use of Project Loom virtual threads

easye:
Thanks to no-defun-allowed for clueing me into that this would be
somewhat easily done.

When the underlying JVM supports virtual threads, :VIRTUAL-THREADS
will be present in CL:&ast;FEATURES&ast;.

The special variable THREADS:&ast;THREADING-MODEL&ast; contains the type of
threads being spawned by the implementation: it will be :NATIVE for
the default, which is usually a one-to-one mapping from each Java
thread to a native thread, and will be :VIRTUAL if using the
lightweight threads present in Project Loom JVMs.

Tested on "OpenJDK_64-Bit_Server_VM-Oracle_Corporation-16-loom+9-316".

Netbeans 12.2 has problems running the debugger with this change due
to some sort of bad interaction with the JDWP protocol.

Project Loom documentation of java.lang.Thread
<https://download.java.net/java/early_access/loom/docs/api/java.base/java/lang/Thread.html>

Overview of Project Loom
<https://cr.openjdk.java.net/~rpressler/loom/loom/sol1_part1.html#migration-from-threads-to-virtual-threads>

Early Access binaries for openjdk16-loom
<https://jdk.java.net/loom/>

### DIFF
--- a/src/org/armedbear/lisp/Lisp.java
+++ b/src/org/armedbear/lisp/Lisp.java
@@ -2429,6 +2429,22 @@ public final class Lisp
             platformVersion = javaVersion;
           }
       }
+      // We wish to declare an integer Java version, but
+      // specialized builds can prefix further information in the
+      // java.version property
+      try {
+	Integer.parseInt(javaVersion);
+      } catch (NumberFormatException e) {
+	for (int i = 0; i < javaVersion.length(); i++) {
+	  char c = javaVersion.charAt(i); // Unicode?
+	  if (!Character.isDigit(c)) {
+	    // Push the non-conforming keyword for completeness
+	    featureList.push(internKeyword("JAVA-" + javaVersion));
+	    platformVersion = javaVersion.substring(0, i);
+	    break;
+	  }
+	}
+      }
       featureList = featureList.push(internKeyword("JAVA-" + platformVersion));
     }
 
--- a/src/org/armedbear/lisp/Lisp.java
+++ b/src/org/armedbear/lisp/Lisp.java
@@ -39,6 +39,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.StringReader;
+import java.lang.reflect.Method;
 import java.math.BigInteger;
 import java.net.URL;
 import java.nio.charset.Charset;
@@ -2485,6 +2486,13 @@ public final class Lisp
         featureList = featureList.push(internKeyword(osArch));
       }
     }
+
+    // Available Threading models
+
+    if (LispThread.virtualThreadingAvailable()) {
+      featureList = featureList.push(internKeyword("VIRTUAL-THREADS"));
+    }
+    
     Symbol.FEATURES.initializeSpecial(featureList);
   }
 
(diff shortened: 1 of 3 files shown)


mdbergmann:
Hello.

Just my two cents.
Is it possible to easily decide which kind of thread to create by the extended API?
I could imagine that since the Java API allows to easy create both by just a different method call, that this should should be reflected in ABCL as well because a project could have a need for both at the same time.

easye:
> Is it possible to easily decide which kind of thread to create by the extended API?

With this patch, one can set THREADS:&ast;THREADING-MODEL&ast; to either`:virtual` or `:native` to switch between creating virtual or native threads.  I have tested virtual threads only lightly by ensuring that systems like LPARALLEL and BORDEAUX-THREADS continue to pass their test suites.  I haven't explored switching threading models at all.  

What sort of "extended API" do you mean?  One that we define, or the one that Project Loom exposes?

mdbergmann:
> What sort of "extended API" do you mean? One that we define, or the one that Project Loom exposes?

I think that the new Java API of Project Loom should be passed through to the Lisp user via LispThread (or the ABCL thread primitive) to allow a similar functionality as that of Java in a simple way.
Switching the *THREADING-MODEL* is the thing that must be used for Bordeax-Threads, but is maybe too cumbersome to be used as an API.
So I'd imagine something like `make-virtual-thread` next to `make-thread`, which could maybe just setup a dynamic binding for *THREADING-MODEL* to `:virtual`.

Just an idea. I'm not very familiar with the internals of ABCL. So, never mind.

no-defun-allowed:
I think having `make-thread` use a threading model variable is a good idea; then any libraries you use wouldn't have to be updated to use `make-virtual-thread` instead of `make-thread`, and you could just bind `*threading-model*` to have everything use your threading model.

For context, my attempt at "porting" ABCL to Loom was part of a discussion I had with @Gnuxie about how to write async code the best way, and we had decided green threads would be better than anything else, as the synchronous code out there would "just work" on a green threaded implementation. `make-virtual-thread` would produce a similar problem, though it would be much easier to fix, as it's just replacing a function name, and not modifying the structure of programs in any other way.

mdbergmann:
> and you could just bind *threading-model* to have everything use your threading model.

Of course, as this allows backwards compatibility.

> "just work" on a green threaded implementation. make-virtual-thread would produce a similar problem, though it would be much easier to fix, as it's just replacing a function name, and not modifying the structure of programs in any other way.

Java 'allows' you to use both, OS threads and virtual threads, even though an `ExecuterService`. The reason is that an application is likely to use both. OS threads have their strength and 'green-threads' have their strength, too.
If you need raw computing power, then use an OS thread, if you need I/O, then use virtual threads.

Having said that, the `*threading-model*` allows both on the Lisp side. My point was just to ease the way of creating a virtual thread.

mdbergmann:
> Having said that, the `*threading-model*` allows both on the Lisp side. My point was just to ease the way of creating a virtual thread.

Or differently said, to make what is being created more explicit than _only_ trough a global variable.

no-defun-allowed:
Okay, I must have misread, sorry. Having a `make-virtual-thread` and `make-os-thread` somewhere would definitely be nicer than `(let ((*threading-model* :os)) (bt:make-thread ...))`, though defaulting to `*threading-model*` for `make-thread` would work most of the time. Or it could be `(make-thread <function> :threading-model :os)`, where `:threading-model` defaults to the value of `*threading-model*`? Not that the exact syntax matters, just that there is an straight-forward way to choose.

PR #365, opened 2021-01-16
Trac/ticket/479/mevenson/20210116a

easye:


easye:
Unfortunately, this breaks something in loading `osicat` (and perhaps
others) as `cl:code-char` starts complaining about 

```
The value * is not of type FIXNUM.
```

PR #367, opened 2021-01-29
Fix printing of `RANDOM-STATE`.

Ferada:
Not a particularly nice byte vector representation though, but it gets the job done:

```
CL-USER> (write (make-random-state))
#<RANDOM-STATE {301D7DA5}>

CL-USER> (write (make-random-state) :readably T)
#.(SYSTEM::READ-RANDOM-STATE #(-84 -19 ... -69 39 120))

CL-USER> (setf *read-eval* NIL)
NIL
CL-USER> (let ((*read-eval* NIL)) (write (make-random-state) :readably T))
#<RANDOM-STATE {3B023252}> cannot be printed readably.
   [Condition of type PRINT-NOT-READABLE]
...
```

[From CLHS](http://clhs.lisp.se/Body/v_pr_rda.htm):
> If `*read-eval*` is false and `*print-readably*` is true, any such method that would output a reference to the ``#.'' reader macro will either output something else or will signal an error (as described above).

Based on that this should be conforming and users will not be able to print a random state if they don't have `*read-eval*` on. Alternatively a different reader implementation would have to be done, e.g. via `#S` like the example in the documentation.

### DIFF
--- a/src/org/armedbear/lisp/RandomState.java
+++ b/src/org/armedbear/lisp/RandomState.java
@@ -68,6 +68,29 @@ public final class RandomState extends LispObject
         }
     }
 
+    public RandomState(SimpleVector v)
+    {
+        int length = v.capacity;
+        byte[] bytes = new byte[length];
+        for (int i = 0; i < length; ++i) {
+            LispObject obj = v.data[i];
+            if (obj instanceof Fixnum) {
+                bytes[i] = (byte)((Fixnum)obj).value;
+            } else {
+                error(type_error(obj, Symbol.FIXNUM));
+            }
+        }
+        try {
+            ByteArrayInputStream byteIn = new ByteArrayInputStream(bytes);
+            ObjectInputStream in = new ObjectInputStream(byteIn);
+            random = (Random) in.readObject();
+            in.close();
+        }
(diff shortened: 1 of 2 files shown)


easye (commented):
One can add docstrings via either

1) use the Java @DocString  annotation e.g. <https://github.com/armedbear/abcl/blob/master/src/org/armedbear/lisp/Pathname.java#L1976>

2) Use the original convention via `###` occurring in a comment c.f. <https://github.com/armedbear/abcl/blob/master/src/org/armedbear/lisp/Primitives.java#L446>

easye:
`@DocString` is preferred, but can be hard to get a Java editor to format it nicely.

Ferada:
> One can add docstrings via either

Okay, I'll add one.

easye:
Thanks for the patch!

Arrgh!  The meta-data *still* doesn't make it SVN…  my apologies.  If you add your name to your commit message like 
```
This patch does X
(Olof)

Longer comment
``` 
It will make it through, although it will still have me as the committer in SVN.  Again my apologies for this, I'll try to figure out how to fix soon.
<https://abcl.org/trac/changeset/15500>

PR #374, opened 2021-05-10
Allow JAVA:JCLASS to work for non-primitive arrays suffixed with "[]"

easye:
This allows constructs like (jclass "java.lang.Integer[]") to work.


easye:
Superseded by #375

PR #379, opened 2021-08-04
Support variadic arguments with more required parameters

eatonphil:
I noticed that under the current systems calls like `(jstatic "asList" "java.util.Arrays" (jnew-array "int" 0))` works because the signature is `java.util.Arrays.asList(T...)` but the system didn't support signatures like `java.nio.file.Path.of(String first, String... more)`. This patch adds support for that and makes the "no method found" message easier to understand when calling `jstatic` with a signature that is not found.

### DIFF
--- a/src/org/armedbear/lisp/Java.java
+++ b/src/org/armedbear/lisp/Java.java
@@ -433,6 +433,7 @@ public final class Java
         }
         try {
             Method m = null;
+            Class c = null;
             LispObject methodRef = args[0];
             List<Method> staticMethods = new ArrayList<Method>();
             String methodName = null;
@@ -446,7 +447,7 @@ public final class Java
                     error(new LispError(methodRef +  "is not a valid reference to a Method"));
                 }
             } else if (methodRef instanceof AbstractString) {
-                Class c = javaClass(args[1]);
+                c = javaClass(args[1]);
                 if (c != null) {
                     methodName = methodRef.getStringValue();
                     Method[] methods = c.getMethods();
@@ -465,8 +466,25 @@ public final class Java
                 m = findMethod(staticMethods.toArray(new Method[staticMethods.size()]),
                                methodName, args, 2);
             }
-            if (m == null)
-                error(new LispError("no such method"));
+            if (m == null) {
+              StringBuilder sb = new StringBuilder("No such static method: ");
+              String className = "<unknown>";
+              if (c != null) {
+                className = c.getName();
+              }
+              sb.append(className);
+              sb.append('.');
+              sb.append(methodName);
+              sb.append('(');
+              for (int i = 2; i < args.length; i++) {
+                LispObject argClass = Symbol.JCLASS_OF.execute(args[i]);
+                sb.append(argClass.princToString());
+                if (i < args.length - 1)
+                  sb.append(',');
+              }
+              sb.append(')');
+              error(new LispError(sb.toString()));
+            }
 
             Object[] methodArgs = new Object[args.length-2];
             Class[] argTypes = m.getParameterTypes();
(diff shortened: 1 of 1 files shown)


easye:
@eatonphil   You wrote in email 

```
I can find the right method with `(jmethod "java.nio.file.Path" "of" "java.lang.String" (jclass-of (jnew-array "java.lang.String" 0)))` but when I call it I get `Java exception 'java.lang.IllegalArgumentException: wrong number of arguments'.
```

I've been trying to come up with an test that shows this failure, but the following test succeeds.  I wonder if your failure message was coming from failing to pass a class to the `jcall`:  even though it is a static method, one needs to to provide an instance when making the call.  

```
     (let* ((parameter
              (jnew-array-from-list "java.lang.String" '("tmp" "passwd")))
            (class
              (jclass "java.nio.file.Path"))
            ;;; N.b. Path.of() only exists starting from openjdk11.  How to test earlier?
            (method
              (jmethod "java.nio.file.Path" "of" "java.lang.String" (jclass-of parameter)))
            (result
              (jcall method class "/chroot/" parameter)))
       result)
```

Could you help me write a failing test that demonstrates what your patch changes?

eatonphil:
```
CL-USER(1): (jstatic "of" "java.nio.file.Path" "foo.bar" (jnew-array-from-list "java.lang.String" '()))
#<THREAD "interpreter" {6C2F5FFA}>: Debugger invoked on condition of type JAVA-EXCEPTION
  Java exception 'java.lang.IllegalArgumentException: wrong number of arguments'.
Restarts:
[1] CL-USER(2): (jmethod "java.nio.file.Path" "of" "java.lang.String" (jclass-of (jnew-array-from-list "java.lang.String" '())))
#<java.lang.reflect.Method public static java.nio.file.Path.... {51C8922}>
[1] CL-USER(3): (setf m (jmethod "java.nio.file.Path" "of" "java.lang.String" (jclass-of (jnew-array-from-list "java.lang.String" '()))))
#<java.lang.reflect.Method public static java.nio.file.Path.... {7DE01E8B}>
[1] CL-USER(4): (jstatic m "foo.bar" (jnew-array-from-list "java.lang.String" '()))
#<THREAD "interpreter" {6C2F5FFA}>: Debugger invoked on condition of type ERROR
  no such method
Restarts:
  0: ABORT     Return to debug level 1.
  1: TOP-LEVEL Return to top level.
[2] CL-USER(5): (jstatic m "java.nio.file.Path" "foo.bar"  (jnew-array-from-list "java.lang.String" '()))
#<THREAD "interpreter" {6C2F5FFA}>: Debugger invoked on condition of type JAVA-EXCEPTION
  Java exception 'java.lang.IllegalArgumentException: wrong number of arguments'.
Restarts:
  0: ABORT     Return to debug level 2.
  1: ABORT     Return to debug level 1.
  2: TOP-LEVEL Return to top level.
```

I made one mistake in there, `CL-USER(4)`, but otherwise does this help?

You can _find_ the method but you cannot _call_ it and that explains why the only code change here is _how_ the args are cast when calling `invoke`.

easye:
> You can _find_ the method but you cannot _call_ it and that explains why the only code change here is _how_ the args are cast when calling `invoke`.

Yes, I can observe the difference in behavior with your patch now.  

Interesting that the test I posted previously in this issue works with `jcall` on the method:  when I replace it with `jstatic` the test fails without your patch.

eatonphil:
> Interesting that the test I posted previously in this issue works with jcall on the method: when I replace it with jstatic the test fails without your patch.

Yes that is surprising. I don't know enough of the code or Java to understand if that's expected.

I just understood from the name that `jcall` was supposed to be for methods and `jstatic` was supposed to be for static methods. Maybe it's not that exact.

alessiostalla:
iirc jstatic was for static *fields* but had this quirky behaviour that you
could invoke *some* static *methods* with it.
This is just from hazy memory.

easye:
Merged as <https://github.com/easye/abcl/commit/f74f177fe666c8bb84c00c965bce649b66b55dda>

eatonphil:
Thanks! How does that fork relate to this repo?

easye:
> How does that fork relate to this repo?

Which fork?  <https://github.com/easye/abcl> is my "personal" fork which I use to make pull requests against <https://github.com/armedbear/abcl> if that is what you are asking about.

eatonphil:
Gotcha yeah I was just confused why the PR against this repo was getting closed and only merged into your fork. But if that's a staging fork then cool!

PR #386, opened 2021-08-25
ci: move to Adoptium for openjdk download

easye:
The CIs now build and test with openjdk8, openjdk11, and openjdk16.


easye:
Superseded by <https://github.com/armedbear/abcl/pull/391>

PR #387, opened 2021-09-06
Remove confusing comment in example

Uthar:
Seems to have been copied by mistake from lisp-to-java/Main.java


easye:
Thanks for the patch!

PR #389, opened 2021-09-17
Fix stream-error-stream, make it and type-error-datum throw type-errors

### LINKED ISSUE #388: stream-error-stream fails on subconditions to stream-error
Title says it all:
```lisp
(define-condition foo-error (stream-error)
  ())
(make-condition 'foo-error :stream (make-string-input-stream "foobar"))
(typep * 'stream-error) ;; => T
(stream-error-stream **)
#<THREAD "interpreter" {119C150C}>: Debugger invoked on condition of type TYPE-ERROR
  The value #<FOO-ERROR {2075326}> is not of type STREAM-ERROR.
Restarts:
```

I can see this fault revealing itself in scenarios where libraries (like parsers) subclass `stream-error`, or in cases where code handling any `stream-error` accesses the error's stream.

Interesting to note, @informatimago found that `type-error` doesn't have this same bug with `type-error-datum`, which seems to imply it's limited to `stream-error`:
```lisp
(define-condition foo-error (type-error)
  ())
(make-condition 'foo-error :datum 42)
(typep * 'type-error) ;; => T
(type-error-datum **) ;; => 42
```

samuel-hunter:
Resolves #388, the function stream-error-stream fails when
applied to subconditions of stream-error:
```lisp
(define-condition foo-error (stream-error)
  ())
(make-condition 'foo-error :stream (make-string-input-stream "foobar"))
(typep * 'stream-error) ;; => T
(stream-error-stream **)
 <THREAD "interpreter" {119C150C}>: Debugger invoked on condition of type TYPE-ERROR
  The value #<FOO-ERROR {2075326}> is not of type STREAM-ERROR.
Restarts:
```

This change both fixes stream-error-stream and revises type-error-datum by
checking if each arg is typep to its respective condition type. If the typep
fails, the functions signal a type-error. It also adds two unit tests in
condition-tests.lisp to protect against this bug regressing back again.

This change would make both functions conformant to two proposals for
WSCL:
- https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/stream-error-stream-type-error
- https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/type-error-datum-type-error

### DIFF
--- a/src/org/armedbear/lisp/StreamError.java
+++ b/src/org/armedbear/lisp/StreamError.java
@@ -161,9 +161,13 @@ public class StreamError extends LispError
         @Override
         public LispObject execute(LispObject arg)
         {
-            if (arg instanceof StreamError)
-                return ((StreamError)arg).getStream();
-            return type_error(arg, Symbol.STREAM_ERROR);
+            final StandardObject obj;
+            if (arg instanceof StandardObject) {
+                obj = (StandardObject) arg;
+            } else {
+                return type_error(arg, Symbol.STANDARD_OBJECT);
+            }
+            return obj.getInstanceSlotValue(Symbol.STREAM);
         }
     };
 }
--- a/src/org/armedbear/lisp/StreamError.java
+++ b/src/org/armedbear/lisp/StreamError.java
@@ -161,12 +161,11 @@ public class StreamError extends LispError
         @Override
         public LispObject execute(LispObject arg)
         {
-            final StandardObject obj;
-            if (arg instanceof StandardObject) {
-                obj = (StandardObject) arg;
-            } else {
-                return type_error(arg, Symbol.STANDARD_OBJECT);
+            if (arg.typep(Symbol.STREAM_ERROR) == NIL) {
+                return type_error(arg, Symbol.STREAM_ERROR);
             }
+
+            StandardObject obj = (StandardObject) arg;
             return obj.getInstanceSlotValue(Symbol.STREAM);
         }
     };
--- a/src/org/armedbear/lisp/TypeError.java
+++ b/src/org/armedbear/lisp/TypeError.java
@@ -191,13 +191,11 @@ public class TypeError extends LispError
         @Override
         public LispObject execute(LispObject arg)
         {
-            final StandardObject obj;
-            if (arg instanceof StandardObject) {
-                obj = (StandardObject) arg;
+            if (arg.typep(Symbol.TYPE_ERROR) == NIL) {
+                return type_error(arg, Symbol.TYPE_ERROR);
             }
-            else {
-                return type_error(arg, Symbol.STANDARD_OBJECT);
-            }
+
+            StandardObject obj = (StandardObject) arg;
             return obj.getInstanceSlotValue(Symbol.DATUM);
         }
     };
--- a/test/lisp/abcl/condition-tests.lisp
+++ b/test/lisp/abcl/condition-tests.lisp
@@ -66,6 +66,11 @@
                #+sbcl    'sb-kernel::object)
   nil)
 
+(deftest stream-error.1
+  (let ((c (make-instance 'stream-error :stream (make-string-input-stream "foo"))))
+    (values (read-line (stream-error-stream c))))
+  "foo")
+
 (deftest type-error.1
   (type-error-datum (make-instance 'type-error :datum 42))
   42)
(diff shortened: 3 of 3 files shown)


src/org/armedbear/lisp/StreamError.java
```diff
@@ -161,9 +161,13 @@ public String getMessage()
         @Override
         public LispObject execute(LispObject arg)
         {
-            if (arg instanceof StreamError)
-                return ((StreamError)arg).getStream();
-            return type_error(arg, Symbol.STREAM_ERROR);
+            final StandardObject obj;
```
informatimago, line 164: It should still test whether the arg is a stream-error or a subclass, and signal a type-error otherwise. Here, it will signal an error about an inexistant slot instead.

informatimago (changes requested):

src/org/armedbear/lisp/StreamError.java
samuel-hunter: Right -- which is the same error that `type-error-datum` returns on a non-`type-error`:
```lisp
CL-USER(1): (make-condition 'condition)
#<CONDITION {7D9FBAFD}>
CL-USER(2): (type-error-datum *)
#<THREAD "interpreter" {657A43BD}>: Debugger invoked on condition of type SIMPLE-ERROR
  The slot SYSTEM::DATUM is missing from the class #<STANDARD-CLASS CONDITION {57A2A10B}>.

CL-USER(4): (stream-error-stream *)
#<THREAD "interpreter" {657A43BD}>: Debugger invoked on condition of type SIMPLE-ERROR
  The slot STREAM is missing from the class #<STANDARD-CLASS CONDITION {57A2A10B}>.
```

For the record, SBCL appears to handle this the same way (while they strictly check the object is type condition, it doesn't check whether it's typep to stream-error or type-error):
```lisp
(make-condition 'condition)
#<CONDITION {1004CB53D3}>
(type-error-datum *)

debugger invoked on a SIMPLE-ERROR in thread
#<THREAD "main thread" RUNNING {1001558173}>:
  There is no slot named SB-KERNEL::DATUM in #<SB-KERNEL::CONDITION-CLASSOID CONDITION>.

(stream-error-stream *)

debugger invoked on a SIMPLE-ERROR in thread
#<THREAD "main thread" RUNNING {1001558173}>:
  There is no slot named STREAM in #<SB-KERNEL::CONDITION-CLASSOID CONDITION>.
```

In any case, an error is an error -- this should be satisfactory.

yitzchak: It's not specified to be a type-error in the ANSI spec (no exceptional situations were listed) but if will probably be in [WSCL](https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/stream-error-stream-type-error) for what it is worth.

samuel-hunter: That's fair -- it looks like WSCL has a mirror proposal for type-error-datum too, so I'll make a separate commit to change the error for both functions.

yitzchak: Awesome.

I have PRs prepped to update the current practices for stream-error-stream and others in WSCL. I'll put them on pause until your PR gets resolved.

samuel-hunter: I've added the changes, thanks for the feedback!

```lisp
CL-USER(3): (make-instance 'standard-object)
#<STANDARD-OBJECT {659BC8F9}>
CL-USER(4): (stream-error-stream *)
#<THREAD "interpreter" {26AA632D}>: Debugger invoked on condition of type TYPE-ERROR
  The value #<STANDARD-OBJECT {659BC8F9}> is not of type STREAM-ERROR.

CL-USER(6): (type-error-datum *)
#<THREAD "interpreter" {26AA632D}>: Debugger invoked on condition of type TYPE-ERROR
  The value #<STANDARD-OBJECT {659BC8F9}> is not of type TYPE-ERROR.
```

I'll go ahead and resolve this thread

samuel-hunter:
It looks like there are similar functions that need some love, too:
- `cell-error-name` checks only that they are an instanceof StandardObject, and returns its slot value.
- `file-error-pathname` returns `NIL` for any object that is not directly a `file-error`.
- `package-error-package` signals a type-error on StandardObjects typep to package-error
- `type-error-expected-type` checks only that they are instanceof StandardObject, and returns its slot value.
- `arithmetic-error-operation` and `arithmetic-error-operands` both throw signal type-errors on StandardObjects typep to package-error.

So two other conditions have the same issue that this PR fixes. I think I'll close this PR, and rewrite a patch for all.

@yitzchak On an aside, it looks like neither `file-error-pathname`, `arithmetic-error-operation`, nor `arithmetic-error-operands` have WSCL proposals like the other readers do. This is the first time I've heard of WSCL, but perhaps it's worth stubbing out proposals for these?

PR #390, opened 2021-09-18
Fix and standardize all error reader functions

### LINKED ISSUE #388: stream-error-stream fails on subconditions to stream-error

samuel-hunter:
Fixes #388, replaces #389. The reader function `stream-error-stream`
signals a `type-error` if it's applied to conditions that's typep to
`stream-error`, but whose class is a subtype to `stream-error`:

```lisp
(define-condition broken-error (stream-error) ())
  BROKEN-ERROR
(stream-error-stream (make-condition 'stream-error :stream (make-string-input-stream "foo")))
  #S(SYSTEM::STRING-INPUT-STREAM)
(stream-error-stream (make-condition 'broken-error :stream (make-string-input-stream "foo")))
  #<THREAD "interpreter" {2075326}>: Debugger invoked on condition of type TYPE-ERROR
    The value #<BROKEN-ERROR {75A83798}> is not of type STREAM-ERROR.
```

This fault exists within these reader functions:

- file-error-pathname (almost -- instead of a type-error, the function returns NIL instead.)
- package-error-package
- arithmetic-error-operation
- arithmetic-error-operands

The fault exists because the Java code checks if its argument is
instanceof its respective Java class. So, the method will fail on any
conditions defined in CL that subclass these errors.

Furthermore, some reader functions only check if they're instanceof
StandardObject, which allows any CLOS object to succeed as long as the
slot the primitive method is looking for exists:

- cell-error-name
- type-error-datum
- type-error-expected-type

Finally, some reader functions exist in standalone Java files, while
others reside in its condition's class file, adding to some
organizational clutter.

This change fills two main jobs:

- Standardize all reader functions to the same behavior: If the argument
  is typep to the right condition, return the right slot value.
  Otherwise, signal a type error.
- Move standalone reader Java code to the files of errors they apply to.

As an aside, this change would make ABCL's reader functions conformant
to a wave of proposals for WSCL:

- https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/cell-error-name-type-error
- https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/stream-error-stream-type-error
- https://github.com/s-expressionists/wscl/blob/main/wscl-issues/proposed/type-error-datum-type-error
- And the rest


easye:
Thank you very much for porting these WSCL details to ABCL!

PR #414, opened 2021-11-26
- Proposal to fix https://github.com/armedbear/abcl/issues/407

### LINKED ISSUE #407: Dribble does not work.
An actual session like the below makes an empty file.

```lisp
CL-USER(0): (dribble "dribble")
CL-USER(1): 1
1
CL-USER(2): 2
2
CL-USER(3): 3
3
CL-USER(4): (dribble)
CL-USER(5):
```

Once the debugger is invoked dribble starts its job.
In such cases finishing the dribble signals the `stream-error`.

The actual session example is like below.

```lisp
CL-USER(0): (dribble "dribble")
CL-USER(1): 1
1
CL-USER(2): 2
2
CL-USER(3): (invoke-debugger (make-condition 'error))
#<THREAD "interpreter" {37704A51}>: Debugger invoked on condition of type ERROR
  #<ERROR {3FC040CF}>
Restarts:
[1] CL-USER(4): 0
CL-USER(5): 3
3CL-USER(6): 4
4CL-USER(7): (dribble)
CL-USER(8):
#<THREAD "interpreter" {37704A51}>: Debugger invoked on condition of type STREAM-ERROR
  java.nio.channels.ClosedChannelException
Restarts:
[1] CL-USER(9): 0
CL-USER(10):
```

Dribbled file contents are like below.

```lisp
CL-USER(5): 3
3CL-USER(6): 4
4CL-USER(7): (dribble)
```

alejandrozf:
Fixes #407 

A  fix/improvement to cl:dribble

This version save almost everything to the specified file
creating a new internal REPL for dribbling and also ends smoothly
when _(dribble)_ is called

### DIFF
--- a/src/org/armedbear/lisp/dribble.lisp
+++ b/src/org/armedbear/lisp/dribble.lisp
@@ -71,7 +71,20 @@
 	   (setf *dribble-stream* new-dribble-stream)
 	   (setf *standard-input* new-standard-input)
 	   (setf *standard-output* new-standard-output)
-	   (setf *error-output* new-error-output)))
+	   (setf *error-output* new-error-output)
+           ;; Starting a new internal REPL for dribbling
+           (loop do
+             (format t "~a> " (package-name *package*))
+             (with-simple-restart (abort "Error detected in dribbling")
+               (handler-case
+                 (let ((input (read *standard-input*)))
+                   (print (eval input) *standard-output*)
+                   (terpri)
+                   (when (equal input '(dribble))
+                     (return)))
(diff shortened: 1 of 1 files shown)


easye:
I haven't had a chance to test, but I don't think your improvement implementation of `cl:dribble` would be worse than the previous one.  Thanks!

alejandrozf:
> I haven't had a chance to test, but I don't think your improvement implementation of `cl:dribble` would be worse than the previous one. Thanks!

You are welcome!

PR #434, opened 2022-01-28
autopopulate *available-disassemblers* and autoload chosen disassemblers

alanruttenberg:


easye:
Chicken and egg: such a probe should perhaps be defined for `javap` here, and then modified on the load of the `abcl-introspect` package.

alanruttenberg:
Your call. I think it just works whether abcl-introspect is there or not. If the disassemblers are somewhere else then you have to manually load them anyways.

PR #437, opened 2022-01-28
(funcall (java:jfield org.armedbear.lisp.Function FUNCTION_CLASS_BYTE…

alanruttenberg:
…S) function) vs (java:jcall getFunctionClassBytes classloader class), which didn't work in all cases

### DIFF
--- a/src/org/armedbear/lisp/disassemble.lisp
+++ b/src/org/armedbear/lisp/disassemble.lisp
@@ -32,6 +32,20 @@
 (in-package :system)
 (require :clos)
 
+(eval-when  (:load-toplevel)
+  (defvar *available-disassemblers*
+	(let ((files 
+		(directory (merge-pathnames
+			    "abcl-introspect/*.lisp"
+			    sys::*abcl-contrib*))))
+	  (loop for f in files
+		append
+		(with-open-file (s f :direction :input)
+		  (loop for line = (read-line s nil :eof)
+			until (eq line :eof)
+			when (search "disassemble-class-bytes" line)
+			  do (return (list (intern (string-upcase (pathname-name f)) :keyword)))))))))
+  
 (defvar *disassembler-function* nil
   "The currently used function for CL:DISASSEMBLE.  
 
@@ -63,6 +77,8 @@ SYS:*DISASSEMBLERS*."
   (flet ((sane-disassembler-p (disassembler)
            (and disassembler
                 (fboundp disassembler))))
+    (when (member name *available-disassemblers*)
+	(require name))
     (setf *disassembler-function*
           (if name
               (let ((disassembler (cdr (assoc name *disassemblers*))))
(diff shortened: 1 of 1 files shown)


src/org/armedbear/lisp/disassemble.lisp
```diff
@@ -32,6 +32,20 @@
 (in-package :system)
 (require :clos)
 
+(eval-when  (:load-toplevel)
+  (defvar *available-disassemblers*
+	(let ((files 
+		(directory (merge-pathnames
```
easye, line 38: One cannot assume the presences of a the ABCL-INTROSPECT in the core implementation. Additionally, the presence of the function in a file does not necessarily mean the assembler is available, as it might have unsatisfied Maven load dependencies.

src/org/armedbear/lisp/disassemble.lisp
```diff
@@ -149,7 +165,9 @@ SYS:*DISASSEMBLERS*."
               (if (or (java:jinstance-of-p classloader "org.armedbear.lisp.MemoryClassLoader")
                       (java:jinstance-of-p classloader "org.armedbear.lisp.FaslClassLoader"))
                   (disassemble-bytes 
-                   (java:jcall "getFunctionClassBytes" classloader class))
+		   (funcall (java:jfield "org.armedbear.lisp.Function" "FUNCTION_CLASS_BYTES") function)
```
easye, line 168: Why this  better than calling `getFunctionClassBytes` on the classloader?  Perhaps it is an additional strategy, rather than a strict replacement?

easye (changes requested):
1.  Rework and move change to probing available disassemblers to a separate commit. 

2.  Explain/change why calling `getFunctionClassBytes()` should be modified.

src/org/armedbear/lisp/disassemble.lisp
alanruttenberg: getFunctionClassBytes didn't work sometimes and I couldn't figure out why, but this works. Should be tracked down. By someone else ;-)

easye: So, we should modify by adding the strategy to an `(or (ignore-errors [CASE A]) (ignore-errors …` clause as it will break even less often…

easye:
Contains changes from [#43](https://github.com/armedbear/abcl/pull/434)

src/org/armedbear/lisp/disassemble.lisp
alanruttenberg: If there aren't any disassemblers then *available-disassemblers* will be nil. It's a directory probe which simply fails. The require loads the appropriate system, which resolves the maven dependencies.

src/org/armedbear/lisp/disassemble.lisp
alanruttenberg: That's fine with me.

easye:
I separated out the FUNCALL fallback into #462.  

The dependency on the ABCL-INTROSPECT contrib should not be present in the core ANSI code, so if you wish to rework that as part of a load of ABCL-INTROSPECT, I will consider such a patch separately.

PR #439, opened 2022-02-05
Proposal to fix Github issue #177 

### LINKED ISSUE #177: DIVISION-BY-ZERO has ARITHMETIC-ERROR-OPERANDS unbound
```
(let ((condition (handler-case (/ 6 0) (division-by-zero (c) c))))
  (arithmetic-error-operands condition))
```

The above evaluates to `(6 0)` on SBCL, CCL, and ECL, but evaluates to `#<UNBOUND>` on ABCL.

alejandrozf:
Fixes #177 
- Filled slots 'operation' and 'operands' of DivisionByZero condition when it is raised

### DIFF
--- a/src/org/armedbear/lisp/Bignum.java
+++ b/src/org/armedbear/lisp/Bignum.java
@@ -301,7 +301,7 @@ public final class Bignum extends LispInteger
 
   public static BigInteger getValue(LispObject obj)
   {
-          
+
     if (obj instanceof Bignum)
       {
         return ((Bignum)obj).value;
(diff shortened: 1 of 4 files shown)


easye:
Seems to break more ANSI tests, with 64 failing: 

<https://github.com/armedbear/abcl/runs/5080167246?check_suite_focus=true#step:22:8020>

TODO: triage to see which tests are actually regressions.

easye:
> TODO: triage to see which tests are actually regressions.

The "more breakage in ANSI" turns out not to be related to this patch, but is occurring due to the addition of load form ordering tests that have been added to ANSI-TEST <https://gitlab.common-lisp.net/ansi-test/ansi-test/-/commit/62f3c4b595a49ac3ba04b7926d42c7e9fc4d0a67>.  The ABCL CI currently clones the `master` ANSI-TEST branch when it runs, so the exact coverage depends on the state of that branch.

easye:
While the whitespace normalization is a bit noisy, it is more correct than what previously existed which is the current practice, namely to clean things up as code is changed.

PR #447, opened 2022-02-21
Add restart on undefined function called offering same-named functions in different packages

alanruttenberg:
You've called a function forgetting the package prefix and you get an undefined function error with restarts. Include in those restarts the possibility of importing (if there are any) functions with the same name from another package continuing.
Useful when debugging.


easye:
Superseded by <https://github.com/armedbear/abcl/pull/448>

easye:
Superseded by #460

PR #456, opened 2022-02-25
Easye/abcl introspect/20220225b

easye:
@alanruttenberg Couple changes from your version of the inspect locals implementation.  If this passes the ABCL CI, I have a corresponding SLIME patch ready to go to which I will amend its commit message to refer to these changes before merging into SLIME upstream.

1.  Move the tests from abcl/t to abcl/contrib/abcl-introspect/t  as they won't work without abcl-introspect being present.

2.  Restore the inclusion of function arguments in the "locals" SWANK makes inspectable.  With your patch, you were doing either locals or arguments but it is useful to see both.  

3.  Start homing symbols which extend SYSTEM in ABCL-INTROSPECT/SYSTEM, automatically adding them to SYSTEM.


easye:
@alanruttenberg Should I remove the extra debugging print stuff around <https://github.com/armedbear/abcl/pull/456/files#diff-be316331cf8a6414e47b6e8e93e8822244ac93c82443b912789c3e1ba792853dL505> the `*debug-locals*` special as well?

alanruttenberg:
I think the debugging code should stay in for a while in case there's a bug that gets exposed with wider use that I need to debug.

The arguments to the function are shown as locals. 

```
(defun foo (&key barney) (print (env-parts)))
(foo :barney 'purple)

((:lexical-variable barney purple) (:block foo #<org.armedbear.lisp.Primitives$BlockMarker@3e5d6c70>))
```

(this is with the precompiler fix) we discussed.

What's been added back is what the arguments were when the function was called, which will usually be the same thing. I do think it's nice to have that and I was going to put it back in when there was a pretty enough way of distinguishing them. I also plan to mark which are closed-over variables vs bound in the body, when the function is a closure,

That we have access to the initial values is gravy. I don't think we'd have access to the original values if it was a strict evaluator, just what the current values are when we went in the the debugger. There's no real reason to keep them around other than for debugging purposes.

PR #458, opened 2022-02-26
Record source location for slot readers and writers

alanruttenberg:


easye:
Thanks for the patch.

PR #459, opened 2022-02-28
Support for pprint of backquote forms, adapted from SBCL

alanruttenberg:
(pprint '`(b ,a)) -> `(b ,a)

Instead of (system::backq-list 'b a)

### DIFF
--- a/src/org/armedbear/lisp/compile-system.lisp
+++ b/src/org/armedbear/lisp/compile-system.lisp
@@ -312,6 +312,7 @@
       (load (do-compile "print.lisp"))
       (load (do-compile "pprint-dispatch.lisp"))
       (load (do-compile "pprint.lisp"))
+      (load (do-compile "pprint-backquote.lisp"))
       (load (do-compile "format.lisp"))
       (load (do-compile "delete.lisp"))
       (load (do-compile "concatenate.lisp"))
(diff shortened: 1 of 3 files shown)


easye (changes requested):
This commit fails five more ANSI-TESTS--namely PRINT.BACKQUOTE.RANDOM.6
PRINT.BACKQUOTE.RANDOM.7, PRINT.BACKQUOTE.RANDOM.8, PRINT.BACKQUOTE.RANDOM.9, and PRINT.BACKQUOTE.RANDOM.12--which should be investigated.

alanruttenberg:
If you look at .6 it's testing \'\`(,@a ,@b). But if you look at the output it is actually testing (system::backq-append) i.e it's lost the splices. The output is a bunch of printer bindings which the printing is tested under. If you bind those variables and pprint  '`(,@a ,@b) it prints fine. 

I think the problem is in the tests. The code for those tests reads:
```
(randomly-check-readability (list (car x)) :test #'is-similar)
```
but should be
```
(randomly-check-readability x :test #'is-similar)
```
By doing that the test is effectively for nil. But `nil is read a nil because we don't have a quasiquote operator, and the reader is even dropping the quote - it does that for any self-evaluating form.

I submitted an [issue](https://gitlab.common-lisp.net/aruttenberg/ansi-test/-/issues/1) at the ansi-test repo

alanruttenberg:
Oh, and those tests fail even without the pull

PR #464, opened 2022-03-12
Maven version isn't always in manifest, fall back to mvn-version-from-mvn-executable if not.

alanruttenberg:


easye:
More Maven love, great!  It would be nice to have some test coverage under the CI for various ways of having Maven installed.

Issue #465, opened 2022-03-16
clos: subclasses not properly updated when superclass changes


alanruttenberg:
```lisp
(defclass a () ((slot)))
(defclass b (a)())
(slot-boundp (make-instance 'b) 'slot)
->
NIL
```
That's correct. Now add an initform to slot in a
```lisp
(defclass a() ((slot :initform 'something)))
(slot-boundp (make-instance 'a) 'slot)
-> t
```
That's correct too. But
```lisp
(slot-boundp (make-instance 'b) 'slot)
-> NIL
```
Is incorrect. It should be t.

alanruttenberg:
Verified that SBCL has the expected behavior.
There's a comment at the top of std-finalize-inheritence

```lisp
  ;; In case the class is already finalized, return
  ;; immediately, as per AMOP.
```
I don't see that in AMOP. I'm speculating that that's the problem. I ran what seemed to be the relevant ansi tests and don't see a difference. I'm going to put in a pull request and we'll see if there are complaints when tests are run.

alejandrozf:
Hi @alanruttenberg 
it seems this is not happening anymore, I don't know when it was fixed but I tried it and I'm getting `t` in the last form

easye:
Indeed, we adopted @alanruttenberg's proposed fix in <https://github.com/armedbear/abcl/commit/61052659768f5dd36dd965b05c753d6b16a98731>.  

Closing.

PR #466, opened 2022-03-16
See https://github.com/armedbear/abcl/issues/465

alanruttenberg:
The patch lets finalize-inheritence run even if a class is already finalized despite what the ABCL source says the AMOP says.


easye:
TODO:  followup by noting deviance in the User manual

PR #483, opened 2022-04-11
getField->getName 

alanruttenberg:
Must be a change from updated javaparser. No idea how this worked with the patch I submitted. Breaks on #1"System.out"


easye:
Thanks!

PR #488, opened 2022-05-08
- This allows load :fset library by adding the too large sexps to loader

easye:
- Macroexpanding all before write the form to the loader to try follow the CL standard


easye:
Superseded by <https://github.com/armedbear/abcl/pull/490>.

PR #501, opened 2022-09-06
Fix to compilation of some specific test-forms

### LINKED ISSUE #500: fails to compile function with TYPE-ERROR
I have  minimal reproducible example:

````
CL-USER(1): (defun test  (m)
  (let ((l 0))
    (declare (type fixnum l))
    (setq l (cond (m 2) ('tt 0)))
    l))
TEST
CL-USER(2): (compile 'test)
#<THREAD "interpreter" {7384CB02}>: Debugger invoked on condition of type TYPE-ERROR
  The value :INT is not of type (MEMBER NIL :BOOLEAN).
Restarts:
````

If you replace the `'tt` with `T` or remove the type declaration of `l`, there is no problem.

So this looks like to be a bug in compiler.

alejandrozf:
Fixes #500 specifically and similar forms but I suspect the problem is bigger

Things like `(defun test  (m)
  (let ((l 0))
    (declare (type fixnum l))
    (setq l (if m 2))
    l))` still won't work


easye:
In testing this fix <https://github.com/easye/abcl/tree/azf-fix-compilation-test-forms/20221210a>, I cleaned up the whitespace a bit. 

A little puzzled at why one can't fix more forms with a little more elbow grease trial and error over collections of failing test cases.   I tried some patching of the p2 compiler forms, but then had problems in the corresponding length of code in the various global lexicals that the compiler state consisted of that was way too hairy for this poor Bear to understand.  I'll try to write up those results soon.  

Otherwise, as always, thanks for the contribution.

alejandrozf:
Thank you, Mark! 

Yeah, I also tried to fix it more completely but faced errors and issues. It is a complex part.
Hopefully we can figure out soon :)

PR #526, opened 2022-11-28
Ci/openjdk update/20221228b

easye:
Test openjdk19 to support fixing <https://github.com/armedbear/abcl/issues/514>.

PR #529, opened 2022-11-30
Ci/update openjdk/20221230c

easye:
Supersedes #526


easye:
Merged with <https://github.com/armedbear/abcl/pull/530>

PR #542, opened 2023-02-05
Fix to the issue when running compile-file on lambda in function position

### LINKED ISSUE #541: Top level form with lambda in function position won't compile
To reproduce create a file with the form 
```
((lambda(x) (+ x 1)) 1)
```
Compile the file and one gets the error
```
The value (lambda (x) (+ x 1)) is not of type symbol.
```

If one wraps the form in a defun and calls that function then the file compiles. Similarly if it is just wrapped with (let () ...).

alejandrozf:
Fixes #541


easye:
Thanks for the patch!

PR #548, opened 2023-02-06
easye/classloader-abstractions/20230305a

easye:
WITH-CLASSLOADER macro for setting the context classloader. 

Should work, but is "IN-PROGRESS" for suggestions.


easye:
Superseded by <https://github.com/armedbear/abcl/pull/549>

PR #557, opened 2023-02-20
build: autoconfigure with common --add-opens for openjdk17+

easye:
Need at least these for the CL+SSL CFFI linkages to work.

### DIFF
--- a/ci/create-abcl-properties.bash
+++ b/ci/create-abcl-properties.bash
@@ -13,6 +13,7 @@ echo "Configuring for $jdk from <${prop_in}>."
 
 # Unused
 # zgc="-XX:+UnlockExperimentalVMOptions -XX:+UseZGC -Xmx<size> -Xlog:gc"
+opens="--add-opens java.base/sun.nio.ch=ALL-UNNAMED  --add-opens java.base/java.io=ALL-UNNAMED"
 
 abcl_javac_source=1.8
 case $jdk in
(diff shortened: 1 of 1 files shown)


alanruttenberg:
Do people typically launch using a script that has the command line options, or do they do java -jar abcl.jar, in which case they need to remember the right add-opens?

easye:
When I "use" ABCL, I am primarily developing it from source, but of course I am a special case here.  As such, I always invoke ABCL by a wrapper script, and use the automation in the `abcl.wrapper` Ant logic <https://github.com/armedbear/abcl/blob/master/build.xml#L588> to regenerate the script.  To switch between various local Java implementations, I use `jenv` <https://github.com/armedbear/abcl/blob/master/ci/install-jenv.bash> <https://github.com/armedbear/abcl/blob/master/ci/ensure-jenv-is-present.bash>.  

Thinking about how to bring some of this configuration to a end user with just binaries, the first step would be to move the UNIX/Windows bash/batch script generation into ABCL-BUILD.  Then ABCL could output a give configuration for invocation.  

The next step would be probably to use the various debugging protocols to have an initial parent abcl process to invoke a separate JVM process slaving i/o, signals, etc. from the parent.  Then, if one could "passivate" the execution context one could save the world somehow in the child JVM, create a new JVM running with possibly different options, and then restore the continuation's execution in the new child.  Far off indeed, but I think quite possible if we require openjdk17 for such things…

alanruttenberg:
I also use a manifest. Was wondering about people who are perhaps used to java -jar abcl.jar
A [SO question](https://stackoverflow.com/questions/56150697/multiple-add-opens-in-java-jar-manifest) suggests that the add-opens can be added to the jar manifest, so perhaps the build should do that.

easye:
> A [SO question](https://stackoverflow.com/questions/56150697/multiple-add-opens-in-java-jar-manifest) suggests that the add-opens can be added to the jar manifest,

I'd be for that, but as I understood things we would to declare a "Module" for ABCL itself.  I'll revisit this possibility, as it seems conceptually cleaner.

PR #559, opened 2023-02-21
Avoid NIL in simple LOOP from FORMAT directives

yitzchak:
Simple LOOP requires only compound forms. Hence NIL is not permitted. Some FORMAT directives (like newline) return NIL as the form when they have nothing to add to the body. Normally this is fine since BLOCK accepts NIL as a form. On the other hand, when the newline directive is inside of an iteration directive this will produce something like

```lisp
(LOOP (fu) nil (bar))
```

This is probably mostly a spec compliance issue. It does not show up as a bug since ABCL's Loop implementation isn't strict in this sense. Loading an alternate LOOP implementation which is more strict like [Khazern](https://github.com/s-expressionists/Khazern/) will cause FORMAT to fail.

PRs have already been submitted and merged for some of the other CMUCL-based FORMAT implementations like Clasp and ECL.

### DIFF
--- a/src/org/armedbear/lisp/format.lisp
+++ b/src/org/armedbear/lisp/format.lisp
@@ -598,7 +598,8 @@
       (multiple-value-bind (form new-directives)
         (expand-directive (car remaining-directives)
                           (cdr remaining-directives))
-        (push form results)
+        (when form
+          (push form results))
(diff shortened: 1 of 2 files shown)


easye:
Superseded by #560 (to rewrite the commit comment).

PR #507, opened 2022-09-28
First working stepper for ABCL!

alejandrozf:
Hi guys

I just completed a working stepper for ABCL (more details below)

See #264 

I think it can help ABCL users even knowing that is an initial attempt.
I've been starting to working in make it compatible with Sly/Slime but it was not trivial and I couldn't
complete a successful integration for now (if someone can do it would be awesome).

If you have some time, please take a look at it, and try to test it.
An maybe we could merge it if you think is ready enougth to be our first stepper.

It can be of course improved, but IMO it can be seen as a working starting point.

I'm attaching a simple session here.

Happy hacking!

Some characteristics:

- For intepreted code
- For plain REPL only, it is still not working with Sly/Slime
When called inside Sly/Slime it will only show print a message and return the form without any stepping
- ? will print a minimal help
- Can inspect variables and symbols in the current package with 'i'
- 'c' will resume the evaluation until the end without the stepper
- 's' will resume the evaluation until the next form to be analyzed
- case-insensitive when inspecting

[stepper-session.txt](https://github.com/armedbear/abcl/files/9667294/stepper-session.txt)

PR #568, opened 2023-02-26
Working stepper for ABCL as a contrib!

alejandrozf:
This PR has the same features as in #507 , but following suggestions was modified to be a contrib module and use more  Lisp instead of Java

Some characteristics:

- For intepreted code
- For plain REPL only, it is still not working with Sly/Slime When called inside Sly/Slime it will only show print a message and return the form without any stepping
- ? will help a minimal help
- Can inspect variables and symbols in the current package with 'i'
- 'c' will resume the evaluation until the end without the stepper
- 's' will resume the evaluation until the next form to be analyzed
- case-insensitive when inspecting
- 'l' will print the local bindings
- 'q' will skip the current stepping evaluation and return NIL
- 'n' will resume the evaluation to stop only in selected symbols

### DIFF
--- a/contrib/abcl-stepper/README.markdown
+++ b/contrib/abcl-stepper/README.markdown
@@ -1,0 +1,810 @@
+ABCL-STEPPER
+============
+
+ABCL-STEPPER provides a working implementation of an stepper as a replacement for the empty cl:step
+
+Some characteristics:
+
+- For intepreted code, it won't step into compiled functions
+
+- For plain REPL only, it is still not working with Sly/Slime. When called inside Sly/Slime main REPL it will only show print a message and return the form without any stepping
+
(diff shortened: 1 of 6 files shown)


alejandrozf:
New features & refinements for the stepper:
- better printing of the current form
- added information about the result of the current form after evaluated
- almost all the code was moved from Java to Lisp
- added options to manipulated the breakpoint in symbols in the middle of the stepping session
- implemented feature to watch variable inside the stepping session
- implemented feature to show the current backtrace at any point of the stepping session
- fixed issue when dealing with FuncallableStandardObject(s)

See the Readme file inside the contrib for more details

alejandrozf:
Added feature step-next that allows to step to the next form without entering into the current form.

See the example added in the Readme for the contrib.

easye:
Superseded via <<https://github.com/armedbear/abcl/pull/596>>

PR #577, opened 2023-04-06
Make `(log x 0)` return 0 (as per the spec)

jpellegrini:
The spec says that

_"If base is zero, log returns zero."_

But ABCL was not checking for zero, so a divide by zero error was generated.

This patch makes it return zero when the base is zero.


easye:
Thanks for the patch!

PR #578, opened 2023-04-30
Return 0 with appropriate inexactness from log

jpellegrini:
A previous commit would return `Fixnum.ZERO` for `(log b n)` if `b` is zero; this commit makes it return `b * n`, that will automatically return the correct zero (`0`, `0.0` or `0.0d0`), according to the inexactness contagion rules.

(Sorry for having missed this in the other PR)


easye:
Thanks for the further attention here.

PR #580, opened 2023-05-04
doc: update notes on packaging

easye:


alanruttenberg:
Minor grammer issues in "The ASDF-JAR contrib provides a mechanism for package ASDF systems and
their recursive dependencies in a jar archive."

"for package" -> "for packaging" or "to package"

Issue #512, opened 2022-10-30
two-way-stream not working for gray streams


Uthar:
recreatable example: make-two-way-stream signals a TYPE-ERROR, IN-STREAM is not of type STREAM
```
(defclass out-stream (trivial-gray-streams:fundamental-character-output-stream) ())

(defclass in-stream (trivial-gray-streams:fundamental-character-input-stream) ())

(make-two-way-stream (make-instance 'in-stream) (make-instance 'out-stream)) ;; BOOM
```

Uthar:
I have the same issue on ABCL for:

- `CL:BROADCAST-STREAM`

```
(make-broadcast-stream (make-instance 'out-stream))
The value #<IN-STREAM {61743FDD}> is not of type STREAM.
```

- `CL:CONCATENATED-STREAM`

```
(make-concatenated-stream (make-instance 'in-stream)
org.armedbear.lisp.StandardObject@6353db8 is not an input stream.
```

- `CL:ECHO-STREAM`

```
(make-echo-stream (make-instance 'in-stream) (make-instance 'out-stream))
The value #<IN-STREAM {47220C0A}> is not of type STREAM.
```

easye:
The basics of accepting a Gray stream as inputs to `make-two-way-stream`, `make-broadcast-stream`, `make-concatenated-stream`, and `make-echo-stream` now work.

PR #582, opened 2023-05-14
Gray streams/GitHub/issue/512

easye:
The stupidest implementation possible for <https://github.com/armedbear/abcl/issues/512>: create an intermediate CLOSProxyStream object, have Lisp.checkStream() return this proxy when passed a CLOS object whose type predicate machinery indicates it is descended from STREAM.

See how this does with the test suites.


easye:
Seems to work!  Fixes <https://github.com/slime/slime/issues/764> with a little further patching of SLIME to load/use the SWANK/GRAY package with ABCL.

Issue #264, opened 2020-08-02
Implementation of STEP


easye:
Request for strategies for implementing `cl:step` for CL implementations in general, and ABCL in specific.

<https://mailman.common-lisp.net/pipermail/armedbear-devel/2020-August/004114.html>

easye:
Via IRC, pjb suggests

There's a conforming stepper at: <https://github.com/informatimago/lisp/tree/master/common-lisp/lisp>

see an example: <https://groups.google.com/forum/#!topic/comp.lang.lisp>/nj3jFxcJYM0

easye:
My workarounds for not using `cl:step` when debugging:
For debugging ABCL interactively under SLIME in lieu of STEP implementation, I visit code in a buffer, copy the top-level to *slime-scratch*, insert BREAK forms, then evaluate the form with C-u slime-compile-defun.  Really suboptimal, but the best route I have found.

That and a lot of use of TRACE to figure out where the problem is occurring.

alanruttenberg:
[copy of email]
I believe I thought the starting point should be in src/org/armedbear/Lisp.java which hosts what looks like an evaluator. Look for

  public static final LispObject eval(final LispObject obj,
                                      final Environment env,
                                      final LispThread thread)

For instance at "else if (obj instanceof Cons)"  it is handling evaluation of a cons, with the first section handling the case where the  first element is a symbol and checking for whether it names a  function, macro, or special form. Below that it handles the case where there's a lambda in the function position. There's a hook for profiling that counts function calls - presumably that would be one place to hook. You'll need to follow it to see how the special forms etc are processed, etc.

If, in slime you execute e.g.  (defun foo (a b &optional c) (and c (+ a b)))
and (inspect #'foo), you'll see how ABCL represents evaluated functions.

easye:
@alejandrozf's interpreted implementation <https://github.com/armedbear/abcl/pull/596> as presented at ELS 2023 is the first candidate <<https://github.com/armedbear/abcl/pull/596>>.  

I'd like to examine the Java-side augmentations this neeed with reference to other `cl:step` abstractions.

Issue #594, opened 2023-06-06
Release abcl-1.9.2


easye:
Major drivers for release:

- [x] fixes for GRAY-STREAMS so that we will work with the next release of SLIME (c.f. <<https://github.com/slime/slime/issues/764>>) which was accomplished by <<https://github.com/armedbear/abcl/pull/586>>.

- [x]  Fix performance regression in implementation of CLOS `mop:specializer` feature  <<https://github.com/armedbear/abcl/issues/587>>

- [x]  Incorporate the interpreted version of `cl:step` <<https://github.com/armedbear/abcl/pull/596>>

Full CHANGES <<https://github.com/armedbear/abcl/blob/b5352f1239413dad12f5ce63851350bc28e90b68/CHANGES>>

easye:
Release available as <https://abcl.org/releases/1.9.2/>

calvinit:
> Release available as https://abcl.org/releases/1.9.2/

[abcl.org](https://abcl.org) is currently inaccessible. May I ask what the issue is? (2025/6/6)

easye:
> > Release available as https://abcl.org/releases/1.9.2/
> 
> [abcl.org](https://abcl.org) is currently inaccessible. May I ask what the issue is? (2025/6/6)

Apparently the TLS certificates need renewing (and abcl.org should be migrated as well).  

Available without TLS as <http://abcl.org> for right now.

PR #596, opened 2023-06-11
ABCL-STEPPER:  implementation of CL:STEP for interpreted code

alejandrozf:
# Working stepper for ABCL now can run in Slime!

This PR was made on top of PR [568](https://github.com/armedbear/abcl/pull/568) and brings to the table the possibility to run the stepper also inside the main REPL in Slime :100: ! (tested on Slime commit 50d4a7b1). 

It is also rebased with the latest changes in ABCL from master branch.

# See also 

Other possible strategies for implementation <<https://github.com/armedbear/abcl/issues/264>>.

### DIFF
--- a/src/org/armedbear/lisp/Lisp.java
+++ b/src/org/armedbear/lisp/Lisp.java
@@ -484,7 +484,7 @@ public final class Lisp
       { threadToInterrupt = thread; }
     else
       { threadToInterrupt = null; }
-    interrupted = b; 
+    interrupted = b;
   }
 
 public static synchronized final void handleInterrupt()
@@ -568,13 +568,26 @@ public static synchronized final void handleInterrupt()
                   if (!sampling)
                     fun.incrementCallCount();
                 // Don't eval args!
-                return fun.execute(((Cons)obj).cdr, env);
+                LispObject stepInSymbolResult = stepInSymbolP(fun, obj);
+                long stepNumberInternal = 0;
+                if (stepInSymbolResult != NIL) {
+                  stepNumber += 1;
+                  stepNumberInternal = stepNumber;
+                  handleStepping(fun, (obj != NIL) ? ((Cons)obj).cdr : obj, env,
+                                 LispInteger.getInstance(stepNumberInternal));
+                }
+                LispObject result = fun.execute(((Cons)obj).cdr, env);
+                if (stepInSymbolResult != NIL) {
+                  printStepValue(stepNumberInternal, result, thread);
+                }
+                setStepCounterCompleted(stepNumberInternal);
+                return result;
               }
             if (fun instanceof MacroObject)
               {
                 try
                   {
-                    thread.envStack.push(new Environment(null,NIL,fun)); 
+                    thread.envStack.push(new Environment(null,NIL,fun));
                     return eval(macroexpand(obj, env, thread), env, thread);}
                 finally
                   {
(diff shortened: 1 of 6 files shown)


easye:
@alejandrozf Congratulations on getting SLIME to work!  I have been spending my recent efforts on deprecating the use of ABCL specialized Slime{Input,Output}Stream <<https://github.com/armedbear/abcl/pull/586>>.  I intend to release abcl-1.9.2 <<https://github.com/armedbear/abcl/issues/594>> "real soon now", for which I will review your stepper for inclusion, so expect feedback early next week.  

best,
Mark

alejandrozf:
@easye 
Thank you, Mark! :) 

Yes, I just tested it for the Slime version v2.28 (commit a4f34714) and it is working fine too

src/org/armedbear/lisp/Lisp.java
```diff
@@ -54,6 +54,72 @@
 
   public static boolean initialized;
 
+  // stepping related attributes
```
easye, line 57: @alejandrozf Please move all the "new" code for the stepper to the end of `Lisp.java`.  Code that changes existing functions is fine to modify "in-place".  

This non-functional change will help in looking through the history of `Lisp.java` with tools that compare lines directly without additional heuristics (like `patch` and `diff`).

src/org/armedbear/lisp/SocketStream.java
```diff
@@ -47,6 +47,10 @@ public SocketStream(Socket socket, Stream in, Stream out)
         this.socket = socket;
     }
 
+    public Socket getJavaSocket() {
```
easye, line 50: I don't have a problem making this interface public, but please separate out into a different pull request.

src/org/armedbear/lisp/Lisp.java
```diff
@@ -2286,6 +2492,136 @@ public LispObject execute(LispObject key, LispObject value)
       }
     };
 
+  // ### %set-stepping-task-on
```
easye, line 2495: A more "Lisp-y" way might be to implement `defsetf-expanders` on various specials, but since these reader/writers are essentially direct toggles of Java booleans, I guess this is fine.

src/org/armedbear/lisp/Lisp.java
```diff
@@ -517,6 +583,73 @@ public static final LispObject eval(LispObject obj)
...
+      funcall(functionPrintStepperStr, new LispObject[] {
+          new SimpleString("step " + stepNumberInternal + " ==> value: " + result.printObject()),
+          Symbol.T
+        },
+        thread);
+    }
+    thread._values = values;
+  }
+
   public static final LispObject eval(final LispObject obj,
```
easye, line 653: Have you done any performance benchmarks?  We are adding additional steps to the Lisp.eval() codepath here, even if we aren't stepping through functions.  From looking at this code, it seems like the only "penalty" is in the boolean compares for `stepInSymbolResult` et. al. but it would be nice to confirm there isn't a generic speed penalty here.

easye (changes requested):
This is more or less ready to commit "as is" modulo reordering the code in `Lisp.java` as far as I am concerned.  

Providing a reference to your 2023 paper would be good as well.

easye:
@alejandrozf Feel free to edit the commit history under this branch (i.e. "git force push") to get the code in a good state.  I can help you squash/edit/reorder commits for you as well.

alejandrozf:
Thank you @easye !

I'll be working on it

Issue #92, opened 2019-01-24
Add Support for Atomic Operators (like CAS, etc.)


Shinmera:
It would be great if ABCL offered support for CAS and other atomic operators in order to allow the implementation of lock-free datastructures and algorithms.

A variety of implementations already offer support for operations like `cas`, `atomic-incf`, `atomic-push`, and `atomic-pop`. For instance, both ECL and SBCL follow the protocol described in the [SBCL manual](http://www.sbcl.org/manual/#Atomic-Operations).

I've accumulated the information for atomic operator support in a new portability library called [Atomics](https://shinmera.github.io/atomics/). For ABCL to be supported, it would need to offer CAS on at least `svref`, `car`, `cdr`, and structure accessor places. Naturally it would be great if it offered more than that still, but those are the most fundamental.

As for how to achieve this: since I don't know anything about ABCL internals I can't say too much, though Java itself offers support for CAS through `sun.misc.Unsafe` and `java.util.concurrent.atomic.*`.

easye:
@Shinmera I'll take a look at what we need to do.  I have some notes from a couple years back that I need to dig up…

Slids:
So I did a little looking.
The obvious thing to do is use varhandles (and not use unsafe as it may go way):
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/invoke/VarHandle.html

Sadly this requires targeting 1.7 and not 1.6

easye:
ABCL 2 will probably target Java 8 (neé 1.8) as the lowest JVM supported, so we can add "true CAS operations" as part of its feature list.  

Thanks for looking into this.

Slids:
I'd be happy to work on this bug but I've no idea when abcl 2 is being release, so if you have some method of notification please drop me a message.

easye:
ABCL 2 is scheduled for September 2020, when I intend to release a version targeting openjdk8

@Slids I guess I should fork ABCL now to give ya a place to add patches.  Stay tuned…

<https://abcl.org/trac/milestone/2.0.0/>

easye:
```
boolean compareAndSet(expectedValue, updateValue)
```
is the basic operation on Java Atomics.  It was introduced in openjdk8, but I wonder if it is possible to synthesize a version in pure Java that would work on openjdk6?  Then we could add a compatibility layer to back-port.

Slids:
The real problem is with calling compare and set on vectors, you have to
look at the type possibilities.

[...]

PR #602, opened 2023-06-28
Explorations in adding atomic compare and swap

easye:
VarHandles were introduced in openjdk9, so the code needs to be conditionalized somehow.  It is currently being developed on openjdk17.

<https://github.com/armedbear/abcl/issues/92>


crmsnbleyd:
is there anything blocking this? maybe this can be added as a package and throw an error on load if jdk is detected as lower than 9?

easye:
Not sure this code really does what we want for CAS, so at a minimum we should develop some tests.  

As for conditionalizing the load we need a proposed implementation that is also tested.

Issue #423, opened 2021-12-29
disassembler problems


alanruttenberg:
I get null pointer exception trying to disassemble anything. I tried this and it seems to work (caveat below).

```
diff --git a/src/org/armedbear/lisp/disassemble.lisp b/src/org/armedbear/lisp/disassemble.lisp
index 03dd1581..2826305a 100644
--- a/src/org/armedbear/lisp/disassemble.lisp
+++ b/src/org/armedbear/lisp/disassemble.lisp
@@ -149,7 +149,9 @@ SYS:*DISASSEMBLERS*."
               (if (or (java:jinstance-of-p classloader "org.armedbear.lisp.MemoryClassLoader")
                       (java:jinstance-of-p classloader "org.armedbear.lisp.FaslClassLoader"))
                   (disassemble-bytes 
-                   (java:jcall "getFunctionClassBytes" classloader class))
+		   (funcall (jss::get-java-field 'org.armedbear.lisp.Function "FUNCTION_CLASS_BYTES") function) 
+                   ;(java:jcall "getFunctionClassBytes" classloader class)
+)
                   (disassemble-bytes
                    (read-byte-array-from-stream
                     (java:jcall-raw
```

```sys:*disassembler*``` is unbound unless ```#'disassemble``` is loaded. I though I tried ```(autoload 'disassemble)``` but that didn't seem to do it, so I did ```(progn #'disassemble)``` before I loaded any of the disassemblers and this fixed the issue.

Ferflower calls ```write-class```, which eventually calls ```write-byte```. The function bytes I get with the diff are apparently ```signed-byte``` and so this fails. For the moment I removed the type check in ```write-byte``` and then fernflower disassembly works.

Procyon doesn't work for me: ```(disassemble #'+)``` -> 
> No such static method: com.strobel.decompiler.Decompiler.decompile([B,com.strobel.decompiler.PlainTextOutput,com.strobel.decompiler.DecompilerSettings)

PR #608, opened 2023-07-10
Fix for procyon decompiler (for cl:disassemble)

alejandrozf:
- This makes procyon disassembler work in ABCL, internally it is using a temp file for the byte array received, then the temp file is passed to the decompiler.
- Updated decompiler status info

See #423

### DIFF
--- a/contrib/abcl-introspect/procyon.lisp
+++ b/contrib/abcl-introspect/procyon.lisp
@@ -4,9 +4,22 @@
      #:disassemble-class-bytes))
 (in-package :abcl-introspect/jvm/tools/procyon)
 
+
+(defun bytes->temp-file (bytes)
+  (let* ((create-temp-file-method
+           (java:jmethod "java.io.File" "createTempFile"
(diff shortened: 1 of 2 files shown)


contrib/abcl-introspect/procyon.lisp
```diff
@@ -4,9 +4,22 @@
      #:disassemble-class-bytes))
 (in-package :abcl-introspect/jvm/tools/procyon)
 
+
+(defun bytes->temp-file (bytes)
+  (let* ((create-temp-file-method
+           (java:jmethod "java.io.File" "createTempFile"
```
easye, line 10: @alejandrozf Note that there already exists a `ext:make-temp-file` interface to File.createTempFile() which could have been re-used.  Not a huge problem at all:  just pointing this out for next time.

easye:
Thanks for the work here.

contrib/abcl-introspect/procyon.lisp
alejandrozf: Oh. I see. Thank you @easye

Issue #612, opened 2023-08-06
Circular lists (maybe)


mnepveu:
Thanks for your answers to my previous post.

I have a data reporting application working with SBCL and CCL, which hangs with ABCL, with one CPU usage going to 100%. The culprit seems to be the "mapcar" call below.

```
(setq *print-circle* t)

(defun number-of-items (rtype datefrom dateto)
  10)

(defun circularlist (items)
  (setf (cdr (last items)) items)
  items)

(mapcar #'number-of-items '("abîmé" "administratif" "autre" "client insatisfait" "coin défecteux" "couleur / profil" "défectueux" "erreur créaction" "erreur du client" "erreur expédition" "impression" "mesure" "offset" "qualité" "transport" "égratignée")
                          (circularlist (list "2023-07-31"))
                          (circularlist (list "2023-08-07")))
```

SBCL
```
This is SBCL 2.3.7, an implementation of ANSI Common Lisp.
More information about SBCL is available at <http://www.sbcl.org/>.

SBCL is free software, provided as is, with absolutely no warranty.
It is mostly in the public domain; some portions are provided under
BSD-style licenses.  See the CREDITS and COPYING files in the
distribution for more information.
* (setq *print-circle* t)
T
* 
(defun number-of-items (rtype datefrom dateto)
  10)
; in: DEFUN NUMBER-OF-ITEMS
;     (SB-INT:NAMED-LAMBDA NUMBER-OF-ITEMS
;         (RTYPE DATEFROM DATETO)
;       (BLOCK NUMBER-OF-ITEMS 10))
; 
; caught STYLE-WARNING:
;   The variable RTYPE is defined but never used.
; 
; caught STYLE-WARNING:
;   The variable DATEFROM is defined but never used.
; 
; caught STYLE-WARNING:
;   The variable DATETO is defined but never used.
; 
; compilation unit finished
;   caught 3 STYLE-WARNING conditions
NUMBER-OF-ITEMS
* 
(defun circularlist (items)
  (setf (cdr (last items)) items)
  items)
CIRCULARLIST
* 
(mapcar #'number-of-items '("abîmé" "administratif" "autre" "client insatisfait" "coin défecteux" "couleur / profil" "défectueux" "erreur créaction" "erreur du client" "erreur expédition" "impression" "mesure" "offset" "qualité" "transport" "égratignée")
                          (circularlist (list "2023-07-31"))
                          (circularlist (list "2023-08-07")))
[...]

Issue #611, opened 2023-08-06
Circular lists


mnepveu:
An infinite loop is entered when attempting to create a circular list.

Environment:
Armed Bear Common Lisp 1.9.3-dev
Java 1.8.0_342 Temurin
OpenJDK 64-Bit Server VM

Examples:

```
(defun circularlist (items)
  (setf (cdr (last items)) items)
  items)

(circularlist (list 1 2 3))
```
Or using Alexandria:

```
(ql:quickload :alexandria)
(alexandria:make-circular-list 1 :initial-element 1)
```

easye:
Both your function `circularlist` and your use `alexandria:make-circular-list` loop forever on SBCL, ECL, and CCL as well, so I think there is a more fundamental issue going on that doesn't work with the semantics of Common Lisp in general.  I must confess that having ALEXANDRIA's inferface not work surprises me as it is a ruggedly battle-tested library.

Closing this issue for now, but if you have more information on what you are trying to do OR if you find code that works on another implementation that works on the Bear, I would be happy to revisit.

alanruttenberg:
Try setting \*print-circle\* to t

```lisp
CL-USER> (setq *print-circle* t)
t
CL-USER> (alexandria:make-circular-list 1 :initial-element 1)
#1=(1 . #1#)
CL-USER> (defun circularlist (items)
  (setf (cdr (last items)) items)
  items)
circularlist
CL-USER> (circularlist (list 1 2 3))
#1=(1 2 3 . #1#)
```

PR #613, opened 2023-08-07
Fix MAPC/MAPCAR for circular lists

easye:
(Alan Ruttenberg)

<https://github.com/mnepveu> reported issues with using circular lists in ABCL with data structures for an application which works in SBCL/CCL but would cause ABCL to hang.

The following test case no longer causes ABCL to hang:

     (setq *print-circle* t)

     (defun number-of-items (rtype datefrom dateto)
       10)

     (defun circularlist (items)
       (setf (cdr (last items)) items)
       items)

     (mapcar #'number-of-items
             '("abîmé" "administratif" "autre" "client insatisfait" "coin défecteux"
               "couleur / profil" "défectueux" "erreur créaction" "erreur du client"
               "erreur expédition" "impression" "mesure" "offset" "qualité" "transport"
               "égratignée")
             (circularlist (list "2023-07-31"))
             (circularlist (list "2023-08-07")))

Implementation from
<https://github.com/armedbear/abcl/issues/612#issuecomment-1667066101>.

See also <https://github.com/armedbear/abcl/issues/611>.

Addresses <https://github.com/armedbear/abcl/issues/612>.

### DIFF
--- a/src/org/armedbear/lisp/Primitives.java
+++ b/src/org/armedbear/lisp/Primitives.java
@@ -2910,36 +2910,43 @@ public final class Primitives {
         }
         @Override
         public LispObject execute(final LispObject[] args)
-
         {
             final int numArgs = args.length;
-            if (numArgs < 2)
+            if (numArgs < 2) {
                 return error(new WrongNumberOfArgumentsException(this, 2, -1));
-            int commonLength = -1;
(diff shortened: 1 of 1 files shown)


easye:
Thanks @alanruttenberg for the implementation.

PR #620, opened 2023-08-31
Implement vector-to-vector `replace` as a primitive

se-mz:
This patch turns vector-to-vector `replace` into a primitive with very fast copies between sequences of the same underlying types (and a slightly faster fallback in Java, although this is primarily for convenience). The current `(setf aref)` loop implementation is very slow, especially for ub8 vectors; code that stitches byte buffers together currently spends a lot of time in `replace` in my experience and a 100MB copy that currently takes ~75s takes ~35ms after this patch.

I'm not too familiar with the ABCL codebase, so some things probably still have to be fixed up. I'm especially unsure whether it's fine to do all the error checking in the primitive and leave the underlying methods unguarded. I'll be in `#abcl` for a while if things are more convenient that way.

### DIFF
--- a/src/org/armedbear/lisp/AbstractVector.java
+++ b/src/org/armedbear/lisp/AbstractVector.java
@@ -307,4 +307,28 @@ public abstract class AbstractVector extends AbstractArray
     {
       return adjustArray(dims[0], displacedTo, displacement);
   }
+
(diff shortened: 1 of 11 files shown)


easye:
Thanks! Reviewing…

PR #621, opened 2023-09-01
Fix to issue when compiling literal quoted forms

### LINKED ISSUE #606: EQness of literals lost
The following definition of FOO has shared structure (see `#1` and `#1#`). But that sharing does not survive compilation.

```
$ cat x.lisp
(defun foo ()
  '(0)
  (let ((x #1='(0))
        (y (second '#1#)))
    (assert (equal x '(0)))
    (assert (equal y '(0)))
    (assert (eq x y))))
```

```
CL-USER(1): (lisp-implementation-version) 
"1.9.1"
"OpenJDK_64-Bit_Server_VM-Debian-17.0.6+10-Debian-1"
"amd64-Linux-6.1.25-1rodete1-amd64"
CL-USER(2): (load (compile-file "x.lisp"))
; Compiling /home/melisgl/own/mgl-pax/x.lisp ...
; (DEFUN FOO ...)
; Wrote /home/melisgl/own/mgl-pax/x.abcl (0.038 seconds)
T
CL-USER(3): (foo)
#<THREAD "interpreter" native {7A9F7D59}>: Debugger invoked on condition of type SIMPLE-ERROR
  The assertion (EQ X Y) failed.
```

alejandrozf:
Fix to issue when compiling literal quoted forms that share parts, fixes #606

### DIFF
--- a/src/org/armedbear/lisp/compiler-pass2.lisp
+++ b/src/org/armedbear/lisp/compiler-pass2.lisp
@@ -1288,6 +1288,61 @@ of the other types."
 4. The function to dispatch serialization to
 5. The type of the field to save the serialized result to")
 
+(defun shared-structure-p (object)
+  "Determines if a literal object shares some structure with previous externalized
+ objects"
+  (let ((circularity-hashtable (make-hash-table :test 'eq)))
+    (labels ((sh-str-helper (obj)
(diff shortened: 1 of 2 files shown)


easye:
Nice work.

Thanks, as always, for looking after the Bear.

PR #624, opened 2023-09-03
Rework SIMPLE-VECTOR implementation using openjdk8 facilities

easye:
(INCOMPLETE) Start of exploration on simplying and improving support for sequences, vectors, and arrays by using generics and functions shipped as part of openjdk8.

Goals:

1)  Simplify class structure with parametertized types

2)  Use openjdk8 methods like Arrays.fill(), sort() et. al.

3)  Reduce memory use for "oversized" buffers

4)  Optimize speed of basic operations

TODO need some benchmarks to test before/after

### DIFF
--- a/src/org/armedbear/lisp/AbstractVector.java
+++ b/src/org/armedbear/lisp/AbstractVector.java
@@ -149,7 +149,7 @@ public abstract class AbstractVector extends AbstractArray
     return index;
   }
 
-  protected void badIndex(int index, int limit)
+  protected LispObject badIndex(int index, int limit)
   {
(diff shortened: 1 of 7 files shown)


easye:
*** Use of generics doesn't seem to help implementation BasicVectorBuffer

1.  One has to "record" the specialization via adding the class to the
   constructor anyways.

2.  The parameterized type can't be used as discriminator type for
   different functions by arguments, as type erasure makes things an
   Object anyways.

Superseded by <https://github.com/armedbear/abcl/pull/624>

PR #632, opened 2023-11-07
Fix to the issue of propagation on existent instances

alejandrozf:
Fixes #630

PR #633, opened 2023-11-11
[azf-fix-630] fixup for @alejandrozf CLOS propagation on existent instances

easye:
Supersedes <https://github.com/armedbear/abcl/pull/632>.

### DIFF
--- a/src/org/armedbear/lisp/clos.lisp
+++ b/src/org/armedbear/lisp/clos.lisp
@@ -3806,6 +3806,22 @@ or T when any keyword is acceptable due to presence of
 
 ;;; make-instances-obsolete
 
+(defmacro breadth-first-search-subclasses (class &body body)
+  `(let ((queue ())
+         visited v)
+       (push ,class visited)
+       (push ,class queue)
+       (loop :while (not (null queue))
+             :do (progn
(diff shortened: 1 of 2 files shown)


easye:
@alejandrozf Thanks for the work, as always.  Your work in <https://github.com/armedbear/abcl/pull/632> didn't merge cleanly so I had to manually edit things to get this merged.  Out of this, I wanted to give you a couple notes that I feel could help improve your contributions to ABCL even more, at least from my perspective.  Please feel free to disagree or argue back if you feel otherwise.

1.  When you have a test for fixing something, it helps to place that test in a separate commit the proceeds the patch that fixes it.  This way it is easy to set up a "before" and "after" situation if one wants to verify that the patch actually fixes the behavior.  

2.  The informal style guideline for commit messages uses the first line of the commit message to summarize the contents of the patch, then followed by a blank line, then as much information, summary and URIs to relevant discussion and context as one feels is necessary to explain the patch.  

Thanks as always for the contribution!

alejandrozf:
Hi @easye

Thanks for the guidelines

They are indeed useful ones

I'll try to follow it for my next contributions :)

PR #636, opened 2023-11-19
Add gray-streams:stream-file-length support

yitzchak:
An extension to Gray streams which allows implementing CL:FILE-LENGTH. Supported by CMUCL, Mezzano, Lispworks,  and SBCL. Recently added to Clasp and ECL.


easye:
Looks good as an implementation as far as I can see so far.  

@yitzchak Curious whether this extension to GRAY-STREAMS is being documented/standardized anywhere?  I'd perhaps like to add a bit of text to the ABCL user manual to note the addition, so looking for something to reference.  And I'd like to understand what else ABCL may be missing that is common in other implementation.  

Thanks for the patch!

yitzchak:
I may submit it too trivial-gray-streams, but I am bit on the fence about that since I have some concerns with how trivial-gray-streams has been implemented. I have started [nontrivial-gray-streams](https://github.com/yitzchak/nontrivial-gray-streams) which is what I think trivial-gray-streams should have been. It includes some initial documentation on the various extensions and the start of a test suite. It also adds features indicating each implementation's support for extensions and capabilities. Still early days, though.

easye:
> I may submit it to trivial-gray-streams […]

Actually, ABCL doesn't have any documentation on its GRAY-STREAMS implementation in our manual at the moment, so adding something quickly won't necessarily be easy.

I'll merge this pull request as-is, follow your progress on <https://github.com/yitzchak/nontrivial-gray-streams>, and deal with some ABCL-side documentation in the future.

yitzchak:
Sounds good. Thanks!

Issue #644, opened 2023-12-03
EVAL of EQUALP across long lists blows up stack


easye:
In triaging why the CALL-ARGUMENTS-LIMIT.4 test fails when invoked under SLIME but works from an unadorned top-level call, I managed to reduce the failure to the following form:

```
  (let ((long-list
          (make-list (expt 2 16) :initial-element 'a)))
    (eval (equalp
           (apply #'list long-list)
           long-list)))
```

This bug is a regression from abcl-1.9.1 in which the ABCL-STEPPER contrib enables interactive stepping through forms by instrumenting the `CL:EVAL` with machinery that should only impact performance when stepping is active.  Apparently this is not the case, so we need to go back through the code to remove side-effects @alejandrozf.

PR #645, opened 2023-12-03
githhub/issues/644/exploration

easye:
# Working changes for <https://github.com/armedbear/abcl/issues/644>

+ differentiate STORAGE-CONDITION causes

+ be more efficient when stepper is invoked

### DIFF
--- a/src/org/armedbear/lisp/Java.java
+++ b/src/org/armedbear/lisp/Java.java
@@ -1384,7 +1384,7 @@ public final class Java
             }
             catch (StackOverflowError oos) {
                 oos.printStackTrace();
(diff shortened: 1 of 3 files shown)


easye (commented):
@alejandrozf

Could you please test that the ABCL-STEPPER still works with this optimization?

I think I made the implementation more efficient when `Lisp.stepping` is `false`, but would appreciate your input here.  Did I miss anything?

alejandrozf:
@easye I just completed an exhaustive test and it seems all still works fine.

I also read the code and it looks good for me too.

Thank you for the optimization!

easye:
Issue is still present, but these patches are useful regardless of how this issue is eventually fixed.

PR #648, opened 2023-12-12
Use STREAMP to determine Gray stream API availability

yitzchak:
Using `(typep obj 'stream)` effectively makes the generic function STREAMP not very useful since all streams need to subclass STREAM, which always returns T for STREAMP. This is an optional part of the Gray stream protocol. Some implementations enforce the requirement that streams must subclass STREAM, i.e. CMUCL, SBCL and CLISP. On the other hand Clasp, ECL, MKCL, Mezzano and CCL all implement the generic STREAMP and don't require that Gray streams subclass STREAM.


easye:
Your reasoning about `streamp` makes sense to me:  I certainly think it is better than the current situation.  Thanks!

PR #651, opened 2023-12-21
Have coerceToPathname use Gray stream PATHNAME generic

yitzchak:
It appears #650 is not quite sufficient for PATHNAME. The spec states that PATHNAME-TYPE, PATHNAME-HOST, etc, should work after [close](https://novaspec.org/cl/f_close). This means that `coerceToPathname` will need to call the generic PATHNAME. To accomplish this I've added `getPathname` to the Stream class.

I've made this a draft because I haven't finished writing tests yet and I haven't verified that TRUENAME works in all cases.

### DIFF
--- a/src/org/armedbear/lisp/Lisp.java
+++ b/src/org/armedbear/lisp/Lisp.java
@@ -2143,12 +2143,10 @@ public static synchronized final void handleInterrupt()
       return (Pathname) arg;
     if (arg instanceof AbstractString)
       return (Pathname)Pathname.create(((AbstractString)arg).toString());
-    if (arg instanceof FileStream)
-      return ((FileStream)arg).getPathname();
-    if (arg instanceof JarStream)
-      return ((JarStream)arg).getPathname();
-    if (arg instanceof URLStream)
-      return ((URLStream)arg).getPathname();
+    Stream s = checkStream(arg);
+    Pathname p = s.getPathname();
+    if (p != null)
+      return p;
     type_error(arg, list(Symbol.OR,
                          Symbol.STRING,
                          Symbol.PATHNAME, Symbol.JAR_PATHNAME, Symbol.URL_PATHNAME,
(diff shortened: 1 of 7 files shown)


src/org/armedbear/lisp/Lisp.java
```diff
@@ -2143,12 +2143,10 @@ public static Pathname coerceToPathname(LispObject arg)
       return (Pathname) arg;
     if (arg instanceof AbstractString)
       return (Pathname)Pathname.create(((AbstractString)arg).toString());
-    if (arg instanceof FileStream)
```
easye, line 2146: From visually inspecting things, I'm surprised that such elision for all the types works.  

The tests didn't complete 'cuz  gitlab.common-lisp.net was wedged for the ANSI-TEST suite.  I've fixed that, re-running tests now.  We'll see.

yitzchak: gitlab.common-lisp.net seems to break a lot. Because Clasp's build system does lots of git cloning and has dependencies on gitlab.common-lisp.net, we've set up a mirror of some the repos. ansi-test is here https://github.com/clasp-developers/ansi-test if you want to use it. It refreshes every day.

yitzchak: Added `@Override` for the various `getPathname` implementations.

yitzchak: I've added tests to nontrivial-gray-streams for the other pathname components. They all pass for this PR.

https://github.com/yitzchak/nontrivial-gray-streams/blob/3275bb649636a962d13cd05498679bd231d538d4/code/test/macros.lisp#L57-L103

easye: Testing looks good wrt. ABCL's PATHNAME specializations (JAR-PATHNAME, URL-PATHNAME).  

Mebbe I will add NONTRIVIAL-GRAY-STREAMS to the CI's testing when I get a chance.

yitzchak:
Thanks @easye!

PR #655, opened 2023-12-30
Gray: Add type checks to default methods to prevent recursion

yitzchak:
Also remove the default method for [FUNDAMENTAL-BINARY-STREAM](https://github.com/yitzchak/nontrivial-gray-streams#fundamental-binary-stream) since this is supposed to be explicitly defined by binary streams.

Removing this method causes an infinite recursion which is the reason for the type checks on the default methods.

### DIFF
--- a/src/org/armedbear/lisp/gray-streams.lisp
+++ b/src/org/armedbear/lisp/gray-streams.lisp
@@ -215,6 +215,15 @@
 (defgeneric gray-pathname (pathspec))
 (defgeneric gray-truename (filespec))
 
+(defun assert-stream (stream)
+  (if (gray-streamp stream)
+      t
+      (error 'type-error :datum stream :expected-type 'stream)))
+
+(defun bug-or-error (stream fun)
+  (assert-stream stream)
+  (error "The stream ~S has no suitable method for ~S." stream fun))
+
 (defmethod gray-close ((stream fundamental-stream) &key abort)
   (declare (ignore abort))
   (setf (stream-open-p stream) nil)
@@ -259,10 +268,6 @@
   'character)
 
 (defclass fundamental-binary-stream (fundamental-stream) ())
-
-(defmethod gray-stream-element-type ((s fundamental-binary-stream))
-  (declare (ignore s))
-  '(unsigned-byte 8))
 
 (defgeneric stream-read-byte (stream))
 (defgeneric stream-write-byte (stream integer))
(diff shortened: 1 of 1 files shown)


src/org/armedbear/lisp/gray-streams.lisp
```diff
@@ -215,6 +215,15 @@
 (defgeneric gray-pathname (pathspec))
 (defgeneric gray-truename (filespec))
 
+(defun assert-stream (stream)
+  (if (gray-streamp stream)
+      t
+      (error 'type-error :datum stream :expected-type 'stream)))
+
+(defun bug-or-error (stream fun)
```
easye, line 223: TODO check if there is an ANSI error we can throw here…

yitzchak: Are you referring to line 221 or to the bug-or-error function?

yitzchak: I think most the "Exceptional situations" require type-errors. I'll scan through them and summarize.

yitzchak: Requires type-error if not stream: [stream-element-type](https://novaspec.org/cl/f_stream-element-type), [input-stream-p](https://novaspec.org/cl/f_input-stream-p), [output-stream-p](https://novaspec.org/cl/f_output-stream-p), [interactive-stream-p](https://novaspec.org/cl/f_interactive-stream-p), or [open-stream-p](https://novaspec.org/cl/f_open-stream-p)

Doesn't specify "Exceptional situations," but the argument is listed as a stream: (close](https://novaspec.org/cl/f_close)

Doesn't specify "Exceptional situations," but the argument is listed as a pathspec, which is generally a string, a pathname or a stream: (pathname](https://novaspec.org/cl/f_pathname) and (truename](https://novaspec.org/cl/f_truename)

yitzchak: Sorry, the conversation thread above got a little confusing. I think I replied while you where editing your original comment.

To summarize my response, I think the type errors are more correct then STREAM-ERROR because the "Exceptional Situation" sections in the above.

easye: I agree with your assessment.  I will merge this when I get a chance.

yitzchak: Thank you!

PR #671, opened 2024-05-01
1.9.3/rc/1

easye:


fosskers:
The version in the title should probably be `1.9.3` yeah?

easye:
> The version in the title should probably be `1.9.3` yeah?

Fixed.  1.9.3 is being "blocked" by problems with CFFI under Apple M1 Silicon.  But CFFI on Apple's arm64 M1 doesn't seem to be working for any (sbcl, ecl, clisp) implementations I can test, therefore this is taking a while to resolve.  To be honest, this isn't a regression to 1.9.2, as apparently CFFI has never worked on arm64 from what I can tell, so I should probably just release what we have.

easye:
Superseded by #677

PR #735, opened 2026-03-05
[BUG FIX] Incomplete TYPEP implementation for single floats.

didierverna:
This fixes the following problems:

(typep 1.0 'number) -> T but (typep 1.0 (find-class 'number) -> NIL
(typep 1.0 'real)   -> T but (typep 1.0 (find-class 'read)   -> NIL

2026-03-05  Didier Verna  <didier@didierverna.net>

	* src/org/armedbear/lisp/SingleFloat.java (typep): Fast-check the type specifier against the REAL and NUMBER built-in classes; not only the symbols.


easye:
Superseded by <https://github.com/armedbear/abcl/pull/736>.

PR #755, opened 2026-08-10
fix: add validate for `:displaced-to`

### LINKED ISSUE #697: Bug: Negative array dimensions permitted when using `:displaced-to` with `make-array`
I discovered this today while unit testing a new library of mine.

This will throw:
```lisp
(make-array -5 :element-type 'character)
```

citing:

> The size specified for this array (-5) is negative.

However, this will happily succeed:

```lisp
(make-array -5 :element-type 'character :displaced-to "hello")
```

This yields an empty string. Each other Lisp implementation I tried throws a condition here. [Looking at the Java](https://github.com/armedbear/abcl/blob/36a243efb86c3667d2f1814ad3a457712f4f92c4/src/org/armedbear/lisp/make_array.java#L83), we see that the dimension isn't checked for negativity when handling `:displaced-to`, but that [it is for other normal cases](https://github.com/armedbear/abcl/blob/36a243efb86c3667d2f1814ad3a457712f4f92c4/src/org/armedbear/lisp/make_array.java#L143).

Cheers.

iwami4438:
- Consolidated the size check logic into a shared function.
- Updated existing code to use the shared function.
- Updated validation for `:displaced-to` to also use the shared function.
- Fixes https://github.com/armedbear/abcl/issues/697


easye (approved):
Looks good; will merge after CI run, and considering other outstanding pull requests.

